PR #13, opened 2026-04-20
chore: reorder workspace deps and normalize formatting

decofe:
Moves internal crate path deps to the bottom under `# crates` and normalizes to `{ path = "..." }` format. Section order is now: alloy → tempo → misc → crates.

Prompted by: zerosnacks


zerosnacks (approved):

mablr (approved):
👌

PR #14, opened 2026-04-20
feat: migrate `foundry-block-explorers`/`foundry-blob-explorers`

mablr:
- Migrate `foundry-block-explorers`/`foundry-blob-explorers` from https://github.com/foundry-rs/block-explorers to `crates/explorers`.
- Updated to rust 2024
- Updated README files


mablr (commented):
This error seems unclear:
```
thread 'verify::test_can_flatten_and_verify_contract_single_file_mainnet' (15788) panicked at crates/explorers/block-explorers/tests/it/verify.rs:47:9:
assertion `left != right` failed: Response { status: "0", message: "NOTOK", result: "Contract source code already verified" }
  left: "NOTOK"
  right: "NOTOK"
```

crates/explorers/block-explorers/tests/it/verify.rs
```diff
@@ -0,0 +1,295 @@
...
+            .runs(500000);
+
+    run_with_client(Chain::mainnet(), |client| async move {
+        let resp = client
+            .submit_contract_verification(&contract)
+            .await
+            .expect("failed to send the request");
+        // `Error!` result means that request was malformatted
+        assert_ne!(resp.result, "Error!", "{resp:?}");
+        assert_ne!(resp.message, "NOTOK", "{resp:?}");
```
zerosnacks:
```suggestion
        // The API returns NOTOK with "Contract source code already verified" for
        // already-verified contracts. To us that's a valid response, not a failure.
        if resp.result != "Contract source code already verified" {
            // `Error!` result means that request was malformatted
            assert_ne!(resp.result, "Error!", "{resp:?}");
            assert_ne!(resp.message, "NOTOK", "{resp:?}");
        }
```

typos.toml
```diff
@@ -17,3 +17,5 @@ extend-exclude = [
 ser = "ser"
 froms = "froms"
 consts = "consts"
+Arguements = "Arguements"
```
zerosnacks, line 20: *Arguments

typos.toml
```diff
@@ -17,3 +17,5 @@ extend-exclude = [
...
+Lod = "Lod"
```
zerosnacks, line 21: This may be a typo

crates/explorers/block-explorers/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+
+[dev-dependencies]
+tempfile.workspace = true
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
+serial_test = "3.2.0"
+tracing-subscriber = { version = "0.3.19", features = ["env-filter", "fmt"] }
+
+[features]
+default = ["rustls"]
+rustls = ["reqwest/rustls"]
```
zerosnacks, line 49: Rather than deprecating openssl flag lets try and see if we can do something like Alloy does:

https://github.com/alloy-rs/alloy/blob/9733a1228516adc51ecb75b2cdeb6a865f4c2291/crates/transport-http/Cargo.toml#L87-L89

typos.toml
mablr: I double checked before adding that, these are false-positives:
- "Arguements" (Etherscan's typo): https://github.com/foundry-rs/foundry-core/blob/346ed9d3e0a7caa5074a8b5673b1b3f321eddfa1/crates/explorers/block-explorers/src/verify.rs#L28-L38 
- "Lod" : https://github.com/foundry-rs/foundry-core/blob/346ed9d3e0a7caa5074a8b5673b1b3f321eddfa1/crates/explorers/block-explorers/src/errors.rs#L83

crates/explorers/block-explorers/Cargo.toml
mablr: Reintroduced `openssl` feat by relaxing `deny.toml` to not ban `openssl` when wrapped in `native-tls`.

Indeed there were no restriction on `openssl` in [block-explorers](https://github.com/foundry-rs/block-explorers) repo

zerosnacks (approved):

PR #19, opened 2026-04-21
fix(wallets): dev origin url without https + 'none' frame-src CSP

mablr:
Fix dev origin url, and hardening CSP w/ frame-src: 'none'.

Tested w/ https://github.com/foundry-rs/foundry-browser-wallet/pull/41

*can be squashed*


zerosnacks (approved):
👍

PR #24, opened 2026-04-22
feat(fork-db): add opt-in zstd cache compression

### LINKED ISSUE #11702: feat(`cache`): add gzip/zstd compression for RPC cache in `foundry-fork-db`
Hi. Would you accept a PR adding compression to revm storage slots (`~/.foundry/cache/rpc/mainnet`) cache? 

I'm working on a project where I want to store revm tracing cache for all the incoming blocks. I'm tracing the last tx from every block, and see cached files up to ~7MB  without any compression. 

Both gzip and zstd reduce size by ~75%, but zstd seems to be significantly faster (20ms vs 170ms on a sample file) and also offers better compression.

Maybe it could be possible to compress cache files, and decompress only for reading? It would significantly reduce disk usage for projects that rely on storing larger amounts of `SharedBackend` cache. Could be an optional flag.

figtracer:
## Summary
- add an opt-in `zstd` feature for cache compression in `foundry-fork-db`
- auto-detect zstd frames on load while keeping plain JSON readable
- log decode failures and add a regression test for zstd data without the feature

Closes foundry-rs/foundry#11702

### DIFF
--- a/crates/fork-db/Cargo.toml
+++ b/crates/fork-db/Cargo.toml
@@ -17,6 +17,10 @@ workspace = true
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
+[features]
+default = []
+zstd = ["dep:zstd"]
+
 [dependencies]
 alloy-chains = { version = "0.2", default-features = false, features = ["serde"] }
 alloy-primitives = { workspace = true, features = ["map"] }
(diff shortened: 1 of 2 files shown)


mablr (approved):
lgtm

zerosnacks (approved):
Nice, lgtm 👍

PR #26, opened 2026-04-23
feat: add OP crates

stevencartavia:
Extract op-alloy-consensus, op-alloy-network, op-alloy-rpc-types, alloy-op-evm, alloy-op-hardforks, and op-revm from upstream into foundry-core.

### DIFF
--- a/crates/op/alloy-op-evm/src/error.rs
+++ b/crates/op/alloy-op-evm/src/error.rs
@@ -1,0 +1,42 @@
+//! Error types for OP EVM execution.
+
+use alloy_evm::InvalidTxError;
+use core::fmt;
+use op_revm::OpTransactionError;
+use revm::context_interface::result::{EVMError, InvalidTransaction};
+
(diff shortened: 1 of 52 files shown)


crates/op/alloy-op-evm/src/error.rs
```diff
@@ -0,0 +1,42 @@
+//! Error types for OP EVM execution.
+
+use alloy_evm::InvalidTxError;
+use core::fmt;
+use op_revm::OpTransactionError;
+use revm::context_interface::result::{EVMError, InvalidTransaction};
+
+/// Newtype wrapper around [`OpTransactionError`] that allows implementing foreign traits.
+#[derive(Debug)]
+pub struct OpTxError(pub OpTransactionError);
```
0xrusowsky: probably useful to impl `Deref`?

zerosnacks:
Closing for now, feature gating seems best direction for the time being

PR #29, opened 2026-04-23
chore: inline compilers inter-crate deps

zerosnacks:
Move internal compilers inter-crate dependencies from workspace `Cargo.toml` to inline `version + path` in each sub-crate's `Cargo.toml`.

This decouples the compilers group's versioning from the workspace root, since each crate group versions independently. Only the cross-group dependency (`foundry-compilers` used by `block-explorers`) remains as a workspace dep.


mablr (approved):
👌

PR #39, opened 2026-04-24
chore: remove `similar-asserts` dependency

zerosnacks:
Remove the `similar-asserts` dependency. It only provided colorized diffs on assertion failure and used the same `PartialEq` comparison as `std::assert_eq!`. All 4 usages were in `#[cfg(test)]` modules and have been replaced with the standard `assert_eq!`.


zerosnacks:
Closes https://github.com/foundry-rs/foundry-core/pull/37

mablr (approved):
nice

figtracer (approved):

PR #51, opened 2026-04-24
feat(block-explorers): add `no_proxy` to `ClientBuilder`

coderHarrii:
## Summary

Mirrors [`reqwest::ClientBuilder::no_proxy`] on `foundry_block_explorers::ClientBuilder`, letting callers disable system proxy detection without having to construct their own `reqwest::Client`.

## Motivation

In sandboxed environments (e.g., Cursor IDE, macOS App Sandbox, restricted shells) `reqwest`'s system proxy lookup via `SCDynamicStore` can crash with:

```
The application panicked (crashed).
Message:  Attempted to create a NULL object.
Location: .../system-configuration-0.6.1/src/dynamic_store.rs:154
```

[foundry-rs/foundry#13155](https://github.com/foundry-rs/foundry/pull/13155) already added a `--no-proxy` flag for the RPC transport. [foundry-rs/foundry#14454](https://github.com/foundry-rs/foundry/pull/14454) does the same for the Etherscan path, but currently has to construct a `reqwest::Client` manually and pipe it in via `with_client(..)`, which forces `foundry-config` to pull in `reqwest` directly.

This PR moves the primitive to the right layer so the foundry-side fix can become a single `.no_proxy()` call.

## Changes

- Add `no_proxy: bool` field to `ClientBuilder`.
- Add `ClientBuilder::no_proxy()` method mirroring `reqwest::ClientBuilder::no_proxy`.
- When a custom client is supplied via `with_client`, `no_proxy` is ignored (user-provided client wins).
- Unit test for `.no_proxy().build()`.

## Follow-up

foundry-rs/foundry#14454 will be updated to call `.no_proxy()` and drop its direct `reqwest` dependency once this ships.

### DIFF
--- a/crates/explorers/block-explorers/src/lib.rs
+++ b/crates/explorers/block-explorers/src/lib.rs
@@ -592,8 +592,7 @@ mod tests {
 
     #[test]
     fn builder_no_proxy_builds() {
-        let client =
-            Client::builder().chain(Chain::mainnet()).unwrap().no_proxy().build().unwrap();
+        let client = Client::builder().chain(Chain::mainnet()).unwrap().no_proxy().build().unwrap();
         assert_eq!(client.etherscan_url.as_str(), "https://etherscan.io/");
     }
 
--- a/crates/explorers/block-explorers/src/lib.rs
+++ b/crates/explorers/block-explorers/src/lib.rs
@@ -256,6 +256,8 @@ pub struct ClientBuilder {
     etherscan_url: Option<Url>,
     /// Path to where ABI files should be cached
     cache: Option<Cache>,
+    /// Whether to disable system proxy detection in the default HTTP client
+    no_proxy: bool,
 }
 
 // === impl ClientBuilder ===
@@ -300,6 +302,19 @@ impl ClientBuilder {
         self
     }
 
+    /// Disables automatic system proxy detection in the default HTTP client.
+    ///
+    /// Mirrors [`reqwest::ClientBuilder::no_proxy`]. Useful in sandboxed
+    /// environments where `reqwest`'s system proxy lookup can panic (for
+    /// example on macOS when `SCDynamicStore` returns NULL).
+    ///
+    /// Has no effect when a custom client is supplied via
+    /// [`with_client`](Self::with_client).
+    pub fn no_proxy(mut self) -> Self {
+        self.no_proxy = true;
+        self
+    }
+
     /// Configures the Etherscan api url
     ///
     /// # Errors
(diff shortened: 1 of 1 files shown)


mablr (commented):
makes sense, ty!

please run `cargo +nightly fmt`

crates/explorers/block-explorers/src/lib.rs
```diff
@@ -330,10 +345,17 @@ impl ClientBuilder {
     ///   - `etherscan_api_url`
     ///   - `etherscan_url`
     pub fn build(self) -> Result<Client> {
-        let ClientBuilder { client, api_key, etherscan_api_url, etherscan_url, cache } = self;
+        let ClientBuilder { client, api_key, etherscan_api_url, etherscan_url, cache, no_proxy } =
+            self;
+
+        let client = match client {
+            Some(c) => c,
+            None if no_proxy => reqwest::Client::builder().no_proxy().build()?,
```
mattsse, line 354: I think this makes sense

mattsse (approved):
pending @mablr

stevencartavia (approved):

DaniPopes (commented):
~~how about a method that takes reqwest clientbuilder directly?~~
we have .with_client - can't this be used?

mablr:
> we have .with_client - can't this be used?

@DaniPopes Indeed, this was submitted here: https://github.com/foundry-rs/foundry/pull/14454

However, it introduces a new `reqwest` dependency on the consumer (foundry) that needs to be synchronized with the version of `foundry-core`. This often becomes messy, which is why I initially proposed the idea to mirror the `reqwest::ClientBuilder` API as we wrap it. In my opinion, this doesn’t hurt.
https://github.com/foundry-rs/foundry/pull/14454#pullrequestreview-4172798634

crates/explorers/block-explorers/src/lib.rs
mablr: Setting `Some(client)` shadows `no_proxy` value, sgtm.

@coderHarrii, I believe it would be beneficial to add a comment about that on the struct field `no_proxy`.

DaniPopes:
there should only be one version anyway, we import reqwest everywhere anyway

crates/explorers/block-explorers/src/lib.rs
coderHarrii: > Setting `Some(client)` shadows `no_proxy` value, sgtm.
> 
> @coderHarrii, I believe it would be beneficial to add a comment about that on the struct field `no_proxy`.

make sense, added!

coderHarrii: > make sense, added!

@mablr

mablr (approved):
g2g if @DaniPopes has no objection

mattsse (approved):

PR #59, opened 2026-05-07
fix(compilers/vyper): support PEP 440 pragmas

figtracer:
Fixes silent loss of Vyper `pragma version` constraints (foundry-rs/foundry#14643).

## Vyper pragma parsing

Two bugs in the Vyper source parser:

1. **Regex never matched the canonical pragma.** `RE_VYPER_VERSION` required `#pragma` with no space, but Vyper also accepts `# pragma version <req>` (which is what e.g. snekmate uses). On top of that the call passed the *outer* match to `VersionReq::parse` instead of the named `version` group, so even when the regex matched, parsing was a no-op.

2. **Pragmas use PEP 440, not semver.** `~=0.5.0a1`, `0.5.0a1`, `==0.5.0` etc. all fail `semver::VersionReq::parse`. Added `parse_vyper_version_req` which first tries plain semver and falls back to a small PEP 440 → semver translation:
   - `~=X.Y.Z` → `>=X.Y.Z, <X.(Y+1).0`
   - `==X.Y.Z` → `=X.Y.Z`
   - `0.5.0a1` / `0.5.0b3` / `0.5.0rc2` → `0.5.0-a1` / `0.5.0-b3` / `0.5.0-rc2`

## Resolver error wording

Surfaced while writing tests for the Vyper fix: `SourceVersionError` and the surrounding error message in `retain_compatible_versions` hardcoded "solc" / "solidity", but the resolver is generic over `Compiler` and is also used for Vyper sources. A failing Vyper pragma constraint therefore produced misleading errors like:

    Encountered invalid solc version in src/Counter.vy: No solc version exists that matches the version requirement: =0.3.10

Replaced the language-specific wording with the neutral "compiler" so the same path produces a correct message for every compiler implementation. The `can_detect_invalid_version` snapshot is updated accordingly. Downstream consumers (foundry, etc.) that snapshot these strings will need to update their assertions when bumping to the next `foundry-compilers` release.


zerosnacks (approved):
lgtm! Have added two mirrored PRs in Foundry to test the behavior in an integrated way:

- https://github.com/foundry-rs/foundry/pull/14664
- https://github.com/foundry-rs/foundry/pull/14667

It occured to me that we don't have something like `svm-rs` for Vyper; we just use the latest compiler version available. Perhaps this is something we can look into in the future if there is demand for it.

PR #61, opened 2026-05-08
feat(wallets): wallet-browser session support 

mablr:
## Summary
Let a single `BrowserSigner` handle multiple consecutive
transaction/signing requests in one session — no webapp reload between
requests.

- Add `GET /api/session` returning `SessionInfo { alive, connected }`
  so the webapp can detect server liveness and stop polling cleanly.
- `stop()` now sets `shutting_down` so the next session poll surfaces
  `alive: false`.
- On disconnect, drain pending tx/signing queues with a synthetic
  "Wallet disconnected" error so callers fail fast instead of waiting
  for the per-request timeout.
- Tests for consecutive txs, disconnect-fail-fast (tx + signing), and
  `/api/session` reporting.
-  Webapp assets refreshed (see: https://github.com/foundry-rs/foundry-browser-wallet/pull/62).

## Testing

In Foundry patch `foundry-wallets` dep with this PR branch.


crates/wallets/src/wallet_browser/signer.rs
```diff
@@ -78,11 +77,8 @@ impl<N: Network> BrowserSigner<N> {
 
         let request = BrowserTransactionRequest { id: Uuid::new_v4(), request: tx_request };
 
-        let server = self.server.lock().await;
```
figtracer, line 81: is it okay to remove the lock here? 

each request starts its timeout as soon as it is enqueued so later browser-wallet requests can time out before the user ever sees them

maybe we should warn?

mablr: > is it okay to remove the lock here?
 
 I removed the mutex on purpose.
 

> each request starts its timeout as soon as it is enqueued so later browser-wallet requests can time out before the user ever sees them
> 
> maybe we should warn?

I think that's the simplest way to impl, as browser is opened not sure the terminal warning is really valuable.

mattsse (approved):
LGTM

figtracer (approved):

crates/wallets/src/wallet_browser/server.rs
figtracer, line 147: can we distinguish explicit user rejection from wallet/send failures, e.g. a separate `Failed` / `WalletError` variant, so CLI output and history do not misclassify infra/relay errors as user rejection?

(a Tempo relay gas error surfaced as “User rejected the request” while testing)

mablr: Good idea

crates/wallets/src/wallet_browser/app/assets/main.js
zerosnacks, line 1: after merging we will do an official release of the foundry-browser-wallet and replace these opaque artifacts; ran an analysis on them and found nothing concerning

zerosnacks (approved):

crates/wallets/src/wallet_browser/server.rs
zerosnacks: Agreed but let's do this in a follow up

PR #67, opened 2026-05-13
feat(wallets): browser/passkey-backed `KeyAuthorization` signing

mablr:
Add browser-wallet bridge support for signing Tempo `KeyAuthorization` payloads with a passkey-rooted account.

Associated to https://github.com/foundry-rs/foundry-browser-wallet/pull/67, https://github.com/foundry-rs/foundry-browser-wallet/pull/75.

- types: `BrowserKeyAuthorizationRequest` / `BrowserKeyAuthorizationResponse`
- queue: `HasId` for the new request type
- state: `key_authorizations` queue + accessors
- handlers/router: `GET /api/key-authorization/request` and `POST /api/key-authorization/response`
- server: `request_key_authorization` with server-side verification of `key_id`, `key_type`, `chain_id` and recovered signer
- signer: `BrowserSigner::sign_key_authorization` high-level entry point
- everything gated behind `tempo` feature


crates/wallets/src/wallet_browser/server.rs
```diff
@@ -203,6 +211,131 @@ impl<N: Network> BrowserWalletServer<N> {
...
+                            Err(e) => {
+                                return Err(BrowserWalletError::ServerError(format!(
+                                    "wallet returned an unrecoverable SignedKeyAuthorization \
+                                     signature: {e}"
+                                )));
+                            }
+                        }
+                    }
+
+                    return Ok(signed);
```
figtracer, line 287: This accepts the wallet's signed payload after only checking key_id/chain_id/key_type. I reproduced a response signed by the connected root that dropped the requested expiry and limits, and the server returned it as ok. Please compare the signed authorization to the requested payload after canonicalization, or reject any mutation that broadens chain or restrictions.

stevencartavia: +1, and worth noting the gap is broader than `expiry`/`limits`: `is_admin` and `account` aren't checked either, so the worst case is escalation to an admin key, not just a weaker one. I reproduced this against `request_keychain_auth` — a response with the same `key_id`/`chain_id`/`key_type` but `into_admin(root)` and no `expiry`, re-signed by the connected root, is accepted as ok.

The Secp256k1 `recover_signer()` check doesn't help: it hashes the wallet-returned authorization, so it only proves root signed whatever came back, not that it matches the request. P256/WebAuthn has no off-chain check at all. Suggest comparing all security-relevant fields against the requested payload (after canonicalization) before accepting.

mablr: Fixed in https://github.com/foundry-rs/foundry-core/pull/67/commits/c7211fca85226489f49e97c8f2383c764619f9ef

crates/wallets/src/wallet_browser/server.rs
```diff
@@ -203,6 +211,110 @@ impl<N: Network> BrowserWalletServer<N> {
...
+                                )));
+                            }
+                            Err(e) => {
+                                return Err(BrowserWalletError::ServerError(format!(
+                                    "wallet returned an unrecoverable SignedKeyAuthorization \
+                                     signature: {e}"
+                                )));
+                            }
+                        }
+                    }
```
stevencartavia: The signature isn't verified for P256/WebAuthn (the passkey case). The gate uses `key_type == Secp256k1`, but `key_type` is the *access key being authorized*, not the root's signing scheme — `recover_signer()` dispatches on the signature variant and verifies all three schemes. So a P256/WebAuthn authorization skips verification and this function can return `Ok` with an unverified/non-root signature (on-chain may still reject later).

Verify unconditionally and drop the gate + stale comment:
```rust
match signed.recover_signer() {
    Ok(recovered) if recovered == root_account => {}
    Ok(recovered) => return Err(BrowserWalletError::ServerError(format!(
        "signed by {recovered} but root account is {root_account}"
    ))),
    Err(e) => return Err(BrowserWalletError::ServerError(format!(
        "invalid SignedKeyAuthorization signature: {e}"
    ))),
}
```
Also update the method doc comment and add `tempo`-gated tests for bad-signature rejection.

mablr: https://github.com/foundry-rs/foundry-core/pull/67/commits/4511b368fa03297957580a053eb05f38a6da97a0

crates/wallets/src/wallet_browser/types.rs
```diff
@@ -110,6 +113,50 @@ pub(crate) struct BrowserSignResponse {
...
+pub struct BrowserKeychainAuthRequest {
+    /// Unique id correlating request and response.
+    pub id: Uuid,
+    /// Root account that must sign the authorization. The wallet UI must
+    /// cross-check this against the connected wallet address before signing.
+    pub root_account: Address,
+    /// The full unsigned `KeyAuthorization` payload (key_id, expiry, limits,
+    /// allowed_calls, ...). Sent so the UI can render a human-readable
+    /// approval card.
+    pub key_authorization: KeyAuthorization,
```
figtracer, line 136: this request can carry `witness`, but the embedded browser-wallet bundle still drops it before `wallet_authorizeAccessKey`. core now correctly rejects mutated returned authorizations, so witness-bearing browser requests cannot succeed until the bundle is regenerated from a browser-wallet fix or unsupported fields are rejected up front.

mablr: https://github.com/foundry-rs/foundry-browser-wallet/pull/77 & https://github.com/foundry-rs/foundry-core/pull/67/commits/6f7686faf9f792561619ec2c4287fcd22d7f3150

figtracer:
The current server-side checks look right now: full returned `KeyAuthorization` equality plus unconditional `recover_signer()` closes the earlier mutation and P256/WebAuthn verification gaps.

Could we add one regression test for a validly signed but mutated authorization, e.g. request has `witness`/`limits`/`account` and the wallet returns a signed payload with one field dropped/changed? That would lock in the security-critical `signed.authorization != key_authorization` rejection.

figtracer (commented):
one note

crates/wallets/src/wallet_browser/signer.rs
```diff
@@ -103,6 +106,33 @@ impl<N: Network> BrowserSigner<N> {
...
+        self.chain_id
+    }
+
+    /// Ask the connected browser wallet to sign a Tempo `KeyAuthorization`.
+    ///
+    /// Cross-checks that the connected wallet is the root account named by
+    /// `key_authorization.chain_id` (when non-zero) and that this signer's
+    /// address matches what we'll send as `root_account`.
+    #[cfg(feature = "tempo")]
+    pub async fn sign_key_authorization(
```
figtracer, line 120: this accepts the full tempo keyauthorization, but the bundled app only forwards key id, chain id, expiry, key type, limits, and scopes to `wallet_authorizeaccesskey`. requests with `witness`, `account`, or `isadmin` will prompt and then fail the digest check because those fields were omitted. please forward/display them or reject them before enqueueing.

mablr: Fixed in 4042c61. The server now rejects T5-only `KeyAuthorization` fields (`witness`, `account`, `is_admin`) before enqueueing a browser signing request, so the browser app will not prompt and then fail the digest check. I also updated the request docs to describe the currently supported field surface and added a regression that verifies a witness-bearing authorization fails without creating a pending browser request.

figtracer (approved):

grandizzy (approved):

PR #71, opened 2026-05-18
fix(ci): fix broken build

decofe:
## Description

Fixes the broken nightly CI build by updating pinned dependency versions.

## Changes

- **alloy**: Bump `alloy-consensus`, `alloy-network`, and all `alloy-signer-*` crates from `2.0.4` → `2.0.5`.
- **alloy**: Bump `alloy-dyn-abi`, `alloy-primitives`, and `alloy-sol-types` from `1.5.7` → `1.6.0`.
- **tempo-primitives**: Switch from the `1.6.0` crates.io release to a git pin (`tempoxyz/tempo` @ `020bc28`) to pick up an unreleased fix needed for compatibility.
- **deny.toml**: Allowlist the `https://github.com/tempoxyz/tempo` git source so `cargo-deny` passes.

## Motivation

The previous version pins were no longer compatible with the latest nightly toolchain / transitive dependency graph, causing CI to fail. These bumps restore a green build.

## Notes

The `tempo-primitives` git pin is temporary and should be reverted to a crates.io release (`> 1.6.0`) once the upstream fix is published.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,7 +127,7 @@ alloy-sol-types = "1.5.7"
 alloy-chains = "0.2.34"
 
 # tempo
-tempo-primitives = { version = "1.6.0", default-features = false, features = [
+tempo-primitives = { version = "1.7.0", default-features = false, features = [
     "std",
     "serde",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,29 +105,29 @@ all = "warn"
 
 [workspace.dependencies]
 # alloy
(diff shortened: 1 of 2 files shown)


mablr (commented):
This is not fixable as-is.

- alloy 2.0.5 depends on alloy-sol-types 1.6.0
- but tempo-primitives 1.7.0 depends indirectly on alloy-sol-types 1.5.7 through tempo-contracts

zerosnacks (approved):

PR #77, opened 2026-05-29
feat(verify): add license type field

mattsse:
Adds `license_type` as a typed `VerifyContract` field for the block explorer verification API and serializes it as Etherscan's documented `licenseType` parameter.

This removes the need for callers to route the parameter through the flattened `other` map, while keeping the field omitted by default so existing verification requests are unchanged.

The change also adds focused serialization coverage for both the set and unset cases.


stevencartavia (approved):

mablr (approved):
Noice

PR #90, opened 2026-06-12
fix(ci): pin time for aws smithy compatibility

figtracer:
## Summary

Pin the `time` crate to `0.3.47` only when the `foundry-wallets` `aws-kms` feature is enabled.

`aws-smithy-types 1.5.0` currently conflicts with `time 0.3.48` on Rust 1.96+/nightly because both crates provide overlapping `From<HourBase>` implementations.

## Notes

A direct AWS SDK bump is not currently viable: newer AWS SDK crates pull in `aws-smithy-async` versions requiring newer `tokio`, while `alloy-signer-turnkey` still depends on `turnkey_client 0.6.2`, which pins `tokio = 1.47.1`.

The `aws_smithy_time_compat` crate-root import exists only to satisfy the workspace `unused_crate_dependencies` lint.


mablr (approved):
Should be fine until next alloy release.

PR #91, opened 2026-06-14
fix(block-explorers): preserve base path in generated explorer URLs

satyakwok:
`block_url` / `address_url` / `transaction_url` / `token_url` build URLs with `etherscan_url.join(...)`. `Url::join` replaces the last path segment when the base has no trailing slash, so an explorer hosted under a sub-path (e.g. `https://example.com/explorer`) produced URLs like `https://example.com/block/1`, dropping `/explorer`.

Normalize the configured explorer URL to end with a trailing slash in `with_url` so joins append instead of replacing. Bare domains and URLs already ending in `/` are unaffected.

Added a regression test (`explorer_url_preserves_base_path`).

### DIFF
--- a/crates/block-explorers/src/lib.rs
+++ b/crates/block-explorers/src/lib.rs
@@ -293,7 +293,13 @@ impl ClientBuilder {
     ///
     /// Fails if the `etherscan_url` is not a valid `Url`
     pub fn with_url(mut self, etherscan_url: impl IntoUrl) -> Result<Self> {
-        self.etherscan_url = Some(into_url(etherscan_url)?);
+        let mut url = into_url(etherscan_url)?;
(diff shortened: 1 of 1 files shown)


mablr (approved):
Makes sense

satyakwok:
Thanks for the review

figtracer (approved):

satyakwok:
@figtracer sorry for the churn — I rebased onto main to bring the branch up to date, not realizing a force-push from a non-write-access account auto-disables auto-merge. The branch is now even with main and both approvals are still in place, so it's ready to re-enable auto-merge or merge directly whenever works.

mablr (approved):

PR #118, opened 2026-07-14
perf(compilers): skip out-of-scope artifact reads

figtracer:
## Summary

- only deserialize cached artifacts whose source belongs to the current compilation graph
- keep full-cache callers unchanged while reducing I/O and JSON work for partial builds
- cover partial compilation with a corrupt out-of-scope artifact, proving unrelated artifacts are not opened
- checked formatting, the focused regression, existing Solidity/Vyper cache tests, and `foundry-compilers` with all features

## Benchmarks

Aave V4 (`271M out`, warm cache), Apple M5 Pro. Measured with `hyperfine --warmup 5 --runs 20`.

| Command | Before | After | Speedup | Artifacts read |
| --- | ---: | ---: | ---: | ---: |
| `forge build src/libraries/math/WadRayMath.sol` | `504.7 ms ± 3.9 ms` | `396.2 ms ± 2.5 ms` | `1.27x` | `664 -> 3` |
| `forge build tests/deployments/batches/AaveV4AuthorityBatch.t.sol` | `540.6 ms ± 6.8 ms` | `443.5 ms ± 1.3 ms` | `1.22x` | `664 -> 235` |

The same scoped reads also reduce a cached filtered test run from `1.147 s ± 0.010 s` to `1.059 s ± 0.004 s` (`1.08x`).

### DIFF
--- a/crates/compilers/crates/compilers/src/cache.rs
+++ b/crates/compilers/crates/compilers/src/cache.rs
@@ -330,11 +330,20 @@ impl<S: CompilerSettings> CompilerCache<S> {
     pub fn read_artifacts<Artifact: DeserializeOwned + Send + Sync>(
         &self,
     ) -> Result<Artifacts<Artifact>> {
+        self.read_artifacts_filtered(|_| true)
+    }
+
+    fn read_artifacts_filtered<Artifact, F>(&self, filter: F) -> Result<Artifacts<Artifact>>
+    where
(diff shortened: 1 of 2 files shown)


mablr (approved):
Makes sense

stevencartavia (approved):

DaniPopes (approved):

PR #121, opened 2026-07-18
feat(wallets): add Touch ID keystore unlock on macOS

### LINKED ISSUE #120: feat(wallets): Touch ID keystore unlock on macOS
Implementation-side tracking issue for foundry-rs/foundry#15815 (Touch ID-protected keystores on macOS). The credential-resolution half of that feature lives in this repo's `foundry-wallets` crate, so it needs to land here first; the follow-up foundry PR only forwards the cargo feature, adds `cast wallet import --touch-id` enrollment UX, and bumps the pinned rev.

Scope for `crates/wallets`, behind an opt-in `touch-id` feature (macOS only, no effect on other targets or default builds):

- Enrollment API that wraps a keystore password with a Secure Enclave-resident P-256 key created behind a Touch ID access-control policy (ECIES: ephemeral ECDH + HKDF-SHA256 + ChaChaPoly), stored in a `.touchid` sidecar next to the keystore JSON, which remains the canonical, portable copy of the key.
- `PendingSigner::unlock` tries the sidecar before the interactive password prompt, so every keystore consumer (`cast send`, `forge script` broadcast, cheatcode wallets) benefits with no per-command changes. Any fallback to the prompt is announced on stderr, never silent.
- A small C-ABI Swift shim compiled by the build script: persisting a Secure Enclave key without keychain entitlements is only possible through CryptoKit's `dataRepresentation`, which has no C surface.

Deliberately uses no macOS Keychain items: biometry-protected items require provisioning-profile entitlements that a plain CLI cannot carry, and file-based keychain ACLs break on every binary upgrade. Full design rationale, platform constraints (Secure Enclave is P-256-only, so signing stays in-process; biometry ACLs and iCloud sync are mutually exclusive), and prior art are in foundry-rs/foundry#15815.

dadadave80:
Adds an opt-in `touch-id` feature to `foundry-wallets` that unlocks keystores with Touch ID on macOS instead of an interactive password prompt. Enrollment wraps the keystore password with a Secure Enclave-resident P-256 key created behind a Touch ID access-control policy (ECIES: ephemeral ECDH + HKDF-SHA256 + ChaChaPoly) and stores it in a `.touchid` sidecar next to the keystore JSON, which remains the canonical, portable copy of the key. `PendingSigner::unlock` tries the sidecar before prompting, so every keystore consumer benefits without per-command changes, and any fallback to the prompt is announced on stderr — a tampered or invalidated sidecar can never silently downgrade to password entry, and a stale wrapped password after a keystore password change falls back with a warning.

The design deliberately avoids macOS Keychain items: biometry-protected items require the data-protection keychain, whose entitlements must be authorized by a provisioning profile embedded in an app-like bundle that a plain CLI cannot carry (Apple TN3137), and file-based keychain "Always Allow" ACLs are bound to the binary's code signature, which would break on every binary upgrade. Persisting a Secure Enclave key without entitlements is only possible through CryptoKit's `dataRepresentation`, which has no C surface, so a small C-ABI Swift shim is compiled by the crate's new build script — only on macOS with the feature enabled; other targets and default builds are unaffected. The Secure Enclave is P-256-only and cannot import keys, so transaction signing still happens in process memory after the password is unwrapped: Touch ID replaces typing the password, not the custody model. Unit tests exercise the sidecar format and a full create/wrap/unwrap Secure Enclave roundtrip non-interactively via a device-only policy (auto-skipped where no enclave exists), with an ignored interactive Touch ID test for manual runs; since CI runners have no Secure Enclave, CI coverage of the enclave path is compile/link-only (the runtime path was verified on hardware), and the clippy job gains a macOS leg so the target-gated code stays lint-checked.

Closes #120. Feature design, platform research, and prior art are in foundry-rs/foundry#15815; a follow-up foundry PR adds `cast wallet import --touch-id` enrollment UX and bumps the pinned rev.

This change was written with AI assistance (Claude).


crates/wallets/src/touch_id/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+        policy,
+        se_key: hex::encode(se_key),
+        sealed_password: hex::encode(sealed_password),
+    };
+    let mut file = fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .mode(0o600)
+        .open(sidecar_path(keystore))?;
```
mablr: This follows and truncates an existing sidecar symlink, so `deployer.touchid -> deployer` destroys the canonical keystore. Write a `0600` temp file in the same directory and atomically rename it; add a symlink regression test.

crates/wallets/src/touch_id/shim.swift
```diff
@@ -0,0 +1,162 @@
...
+    setOut(Data(message.utf8), outPtr, outLen)
+    return 1
+}
+
+private func accessControl(policy: Int32) throws -> SecAccessControl {
+    var flags: SecAccessControlCreateFlags = [.privateKeyUsage]
+    switch policy {
+    case 1: flags.insert(.userPresence)
+    case 2: flags.insert(.biometryCurrentSet)
+    default: break
```
mablr: Unknown policy values fail open to no user presence. Reject unknown values, and keep `DeviceOnly` private/test-only, the public Touch ID API should expose only user-presence/biometric policies.

crates/wallets/src/signer.rs
```diff
@@ -356,6 +356,29 @@ impl PendingSigner {
...
+                                eth_keystore::KeystoreError::MacMismatch,
+                            )) => eprintln!(
+                                "Warning: the Touch ID-stored password no longer decrypts \
+                                     this keystore (was the password changed?); falling back to \
+                                     the password prompt. Re-enroll to refresh it."
+                            ),
+                            Err(e) => return Err(WalletSignerError::Local(e)),
+                        },
+                        // E.g. cancelled prompt, headless session, or invalidated key.
+                        Err(e) => eprintln!(
```
mablr: Cancellation, headless/unavailable auth, invalidation, and malformed sidecars are all downgraded to a password prompt. Return structured FFI errors: cancellation should abort; only recoverable unavailable/stale cases should fall back with a clear message.

.github/workflows/ci.yml
```diff
@@ -90,10 +90,16 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     permissions:
       contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        # macOS covers the target-gated `touch-id` code, which the Linux leg
+        # never compiles.
+        os: ["ubuntu-latest", "macos-latest"]
```
mablr, line 45: One native macOS leg only validates its host architecture, but Foundry ships both `x86_64-apple-darwin` and `aarch64-apple-darwin`. Add target-specific compile/link smoke checks for both, including a final binary link path.

mablr (commented):
Thanks for the feat proposal.

crates/wallets/README.md
```diff
@@ -17,3 +17,4 @@ Supports multiple signer backends:
 - `aws-kms`: [AWS KMS](https://aws.amazon.com/kms/) signer support via [`alloy-signer-aws`](https://docs.rs/alloy-signer-aws).
 - `gcp-kms`: [GCP KMS](https://cloud.google.com/kms) signer support via [`alloy-signer-gcp`](https://docs.rs/alloy-signer-gcp).
 - `turnkey`: [Turnkey](https://www.turnkey.com) signer support via [`alloy-signer-turnkey`](https://docs.rs/alloy-signer-turnkey).
+- `touch-id`: Touch ID (Secure Enclave) keystore unlock on macOS. Compiles a Swift shim at build time and requires the Xcode command line tools; no effect on other targets.
```
mattsse: Could we make this requirement precise? In a clean checkout, `cargo test -p foundry-wallets --features touch-id` reaches the final link and then fails inside Apple `ld` with Command Line Tools 14.3 / Swift 5.8.1, while the new CI coverage uses only `macos-latest`. Please determine and document the minimum supported Xcode/Command Line Tools version here, and add a lowest-supported-toolchain link job (or reject older toolchains early in `build.rs` with a clear error).

mattsse (commented):
One portability concern around the Xcode requirement.

crates/wallets/README.md
dadadave80: Sure, on it

dadadave80: Thanks for the review! Made the requirement precise: the supported build floor is now Xcode/Command Line Tools 15.0.1 (Swift 5.9) — determined empirically by linking both targets across every Xcode still available on hosted runners. Since the macOS 13 images were retired in December, a 14.3/5.8 toolchain can no longer be exercised in CI, and we didn't want to claim support we can't test — it may still work, but it's below the documented floor. Enforcement is both-ends: `build.rs` rejects older toolchains before compiling with an actionable error (probed via `swiftc --version`, so Command Line Tools-only installs are handled), and `touch-id-link` gained an oldest-supported leg that pins `DEVELOPER_DIR` to 15.0.1 and fails loudly if the runner image ever drops it. It's documented in the wallets README. If the `ld` output from the 14.3 build points at something trivially fixable, happy to revisit the floor.

The branch is also merged up with main past the tempo-alloy refactor, with mablr's four review points addressed in the earlier commits. Full suite is green on the [fork](https://github.com/dadadave80/foundry-core/actions/runs/30218893716). Upstream CI here just needs workflow approval when you have a sec.

crates/wallets/src/touch_id/mod.rs
dadadave80: Nice catch, thanks — fixed in [8dce2f8](https://github.com/foundry-rs/foundry-core/pull/121/changes/8dce2f8617f7a98310c7389810bb6f5de6c82d8a). Sidecar writes now go through `write_sidecar()`: a fresh `0600` temp file in the sidecar's directory (via `tempfile`), `sync_all`, then an atomic rename, which replaces a symlink at the sidecar path instead of following it. The regression test `sidecar_write_replaces_symlink_instead_of_following_it` stages the exact attack (`deployer.touchid -> deployer`) and asserts the keystore is untouched and the path now holds a regular `0600` file — it exercises the helper directly, so it also runs on enclave-less CI. Side effects: the write is crash-safe now, and re-enrollment repairs a loosened sidecar mode, so the old `set_permissions` workaround is gone.

crates/wallets/src/touch_id/shim.swift
dadadave80: Fixed in [8dce2f8](https://github.com/foundry-rs/foundry-core/pull/121/changes/8dce2f8617f7a98310c7389810bb6f5de6c82d8a). The shim's switch now accepts exactly the known policy values and throws on anything else, so an unknown value can never mint a key with weaker protection than requested. `Policy::DeviceOnly` is `#[cfg(test)]` — the public API exposes only `UserPresence`/`CurrentBiometry`, the non-interactive CI roundtrip keeps working, and release builds reject a hand-crafted `device-only` sidecar at parse time. (That last bit is belt-and-suspenders — enforcement proper is the enclave key's access control fixed at creation — but our code can no longer produce or accept device-only outside tests.)

crates/wallets/src/signer.rs
dadadave80: Fixed in [8dce2f8](https://github.com/foundry-rs/foundry-core/pull/121/changes/8dce2f8617f7a98310c7389810bb6f5de6c82d8a). The FFI now returns distinct statuses — canceled / unavailable / invalidated / invalid-data — mirrored as `TouchIdError` variants. To make classification reliable, unwrap pre-authenticates with an explicit `LAContext.evaluatePolicy` matching the sidecar's policy (the FFI gained a `policy` param): cancellation and headless/unavailable surface as typed `LAError`s, decode failures before auth are corrupt-sidecar, and an enclave rejection after successful auth is an invalidated key; the authenticated context then satisfies the key's ACL without a second prompt (verified single-prompt on hardware). `unlock` aborts on cancellation and on corrupt/invalidated sidecars (with re-enroll-or-delete guidance in the message) and falls back — always with a stderr warning — only for unavailable auth and the stale-password (`MacMismatch`) case. One deliberate call: unknown Secure Enclave errors also abort (fail closed; the message includes the delete-the-sidecar escape hatch) — happy to move unknowns onto the fallback path instead if you'd prefer.

.github/workflows/ci.yml
dadadave80: Added in [8dce2f8](https://github.com/foundry-rs/foundry-core/pull/121/changes/8dce2f8617f7a98310c7389810bb6f5de6c82d8a) and extended since: a `touch-id-link` job matrixed over `x86_64-apple-darwin` and `aarch64-apple-darwin` running `cargo test -p foundry-wallets --features touch-id --target <t> --no-run`. `--no-run` still performs the final binary link — Swift static lib, runtime search paths, framework autolinking — which `check`/`clippy` never exercise, while the x86_64 binaries simply aren't executed on the arm64 runner. `build.rs` already mapped `CARGO_CFG_TARGET_ARCH` into `swiftc -target`; this job is what finally proves that claim. The follow-up toolchain-floor commits then grew the matrix with an oldest-supported Xcode (15.0.1) dimension, so both targets now link on both the floor and the latest toolchain.

dadadave80: I'll rearrange the touch-id-link ci in a while

.github/workflows/ci.yml
```diff
@@ -108,6 +114,54 @@ jobs:
...
+  # host-arch test leg only ever links one combination. `--no-run` still
+  # performs the final binary link (Swift static lib, runtime search paths,
+  # framework autolinking), which `cargo check`/`clippy` never exercise, while
+  # the x86_64 binaries simply aren't executed on the arm64 runners.
+  #
+  # The oldest leg pins the documented toolchain floor (Xcode 15.0.1,
+  # Swift 5.9 — see `MIN_SWIFT` in crates/wallets/build.rs).
+  # If the runner image drops that Xcode, the pin step fails loudly so the
+  # floor is re-evaluated instead of silently sliding forward.
+  touch-id-link:
```
mattsse, line 71: This job is not included in ci-success.needs, while main requires only ci-success, so failures in this matrix will not block merging. Please add touch-id-link to the aggregate needs list.

mattsse (commented):
LGTM otherwise — pending @mablr.

crates/wallets/src/touch_id/shim.swift
```diff
@@ -0,0 +1,295 @@
...
+    case policyUserPresence: laPolicy = .deviceOwnerAuthentication
+    case policyCurrentBiometry: laPolicy = .deviceOwnerAuthenticationWithBiometrics
+    default:
+        throw ShimError(status: statusFailure, message: "unknown access-control policy \(policy)")
+    }
+
+    var check: NSError?
+    guard context.canEvaluatePolicy(laPolicy, error: &check) else {
+        let message = check?.localizedDescription ?? "authentication is not available"
+        throw ShimError(status: statusUnavailable, message: message)
```
mablr: This collapses every `canEvaluatePolicy` failure into recoverable `Unavailable`, so an invalid-context, internal, or future LA error reaches the password prompt even though only the explicitly whitelisted environmental failures should recover. Please route a non-nil `check` through `classify`, and use `statusFailure` if it is unexpectedly nil.

crates/wallets/build.rs
```diff
@@ -0,0 +1,104 @@
...
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    // Apple Silicon, plus Intel Macs whose T2 chip provides a Secure Enclave.
+    let arch = match env::var("CARGO_CFG_TARGET_ARCH").unwrap().as_str() {
+        "aarch64" => "arm64".to_string(),
+        arch => arch.to_string(),
+    };
+    // Honor the standard deployment-target override, like the `cc` crate does.
+    // CryptoKit's Secure Enclave and HKDF APIs need macOS 11; swiftc rejects
+    // older values with clear availability errors.
+    let deployment = env::var("MACOSX_DEPLOYMENT_TARGET").unwrap_or_else(|_| "11".to_string());
```
mablr: When this environment variable is unset, the Swift object targets macOS 11 but rustc still defaults the final `x86_64-apple-darwin` binary to macOS 10.12. That can produce an executable whose deployment metadata claims support below the Swift code's actual floor. Please align the Rust and Swift deployment policy, and have the link smoke check verify the final Mach-O minimum (or add availability guards if the broader x86 floor must remain unchanged).

.github/workflows/ci.yml
mablr: +1

crates/wallets/src/touch_id/shim.swift
dadadave80: Fixed in 23f54b7, exactly along these lines: a non-nil `check` is now routed through `classify()`, so only the explicitly whitelisted environmental LA codes map to `statusUnavailable` and can reach the password prompt — an invalid context, internal error, or any future LA code lands on `statusFailure` and aborts with the message instead. An unexpectedly nil error is also `statusFailure` now, so nothing reaches the recoverable path by default. A follow-up (4ff72ae) tightened the same boundary further by splitting biometry lockout out of the recoverable set into its own non-recoverable status.

crates/wallets/build.rs
dadadave80: Fixed in [ef88f46](https://github.com/foundry-rs/foundry-core/pull/121/changes/ef88f46cc087fdc6cd853455111ae9a6f0d2ca3b). The Rust toolchain is now the single source of truth: build.rs reads `rustc --print deployment-target --target $TARGET` and passes that same value to swiftc, so the Swift object and the final binary can no longer disagree on the floor. Of your two options we took "align the policy" rather than availability guards: the shim's Swift floor was already macOS 11 (CryptoKit's Secure Enclave and HKDF availability), so guards would only add dead paths for OS versions where the enclave APIs don't exist. The derived value is gated at >= 11.0 with an actionable error, which means an x86_64 build on rustc's 10.12 default (as you noted) now fails fast with instructions instead of emitting an executable whose metadata undersells its real requirement — the README documents that Intel from-source builds set MACOSX_DEPLOYMENT_TARGET=11.0. And the link smoke check verifies the final Mach-O as requested: touch-id-link pins MACOSX_DEPLOYMENT_TARGET=11.0 and asserts `minos` on the linked test executables via `xcrun vtool -show-build`, on both targets.

crates/wallets/src/touch_id/shim.swift
```diff
@@ -0,0 +1,303 @@
...
+    let ns = error as NSError
+    if ns.domain == LAError.errorDomain, let code = LAError.Code(rawValue: ns.code) {
+        switch code {
+        case .userCancel, .appCancel, .systemCancel:
+            return ShimError(status: statusCanceled, message: "authentication was canceled")
+        case .authenticationFailed:
+            return ShimError(status: statusFailure, message: "authentication failed")
+        case .biometryLockout:
+            return ShimError(status: statusLockedOut, message: ns.localizedDescription)
+        case .passcodeNotSet, .biometryNotAvailable, .biometryNotEnrolled, .notInteractive:
```
mablr, line 114: For `CurrentBiometry`, removing all fingerprints reports `biometryNotEnrolled` here. Mapping that to `Unavailable` makes unlock fall back to the password prompt even though `.biometryCurrentSet` invalidated the enrollment. Could `classify` take the policy and map this case to `statusInvalidated`?

dadadave80:
Thank you @mablr @mattsse for the time and rigor across four review rounds. The symlink-truncation catch, the fail-open policy default, pushing the error handling into a real taxonomy instead of a blanket fallback, the toolchain-floor precision, and yesterday's CurrentBiometry invalidation edge each made this materially safer than what was first submitted; the diff between the first commit and today is largely your fingerprints, which feels appropriate for this feature. Thanks as well to the other code owners who've taken a look — review time is the scarcest resource in a project like this, and it's appreciated. Nothing needed from this comment; the latest round is addressed in 6f4ce63.

mablr (approved):
Tested new API whole surface with AI-written foundry fixture. UX is cool.

LGTM, thank you @dadadave80 !

mattsse (approved):

PR #125, opened 2026-07-22
fix(compilers): exclude unrelated re-export imports

stevencartavia:
Resolve named imports to their declaration-owning source units when flattening, preventing unrelated dependencies from intermediary re-export files from being included. Plain and namespace imports remain conservative, and incomplete AST data falls back to the full dependency closure.

Towards foundry-rs/foundry#7239.

### DIFF
--- a/crates/compilers/crates/compilers/src/flatten.rs
+++ b/crates/compilers/crates/compilers/src/flatten.rs
@@ -1,6 +1,6 @@
 use crate::{
     ArtifactOutput, CompilerSettings, Graph, Project, ProjectPathsConfig, SourceParser, Updates,
-    apply_updates,
+    VersionedSourceFiles, apply_updates,
     compilers::{Compiler, ParsedSource},
     filter::MaybeSolData,
     resolver::parse::SolData,
@@ -207,6 +207,15 @@ impl Flattener {
 
         let output = output.compiler_output;
 
+        // AST IDs are only unique within a single compiler invocation. Select the build that
+        // contains the target so IDs from different builds are never mixed.
+        let target_build_id = output
+            .sources
+            .0
+            .get(target)
+            .and_then(|files| files.iter().find(|file| file.source_file.ast.is_some()))
+            .map(|file| file.build_id.clone());
+
         let sources = Source::read_all([target.to_path_buf()])?;
         let graph = Graph::<C::Parser>::resolve_sources(&project.paths, sources)?;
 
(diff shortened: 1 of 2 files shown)


crates/compilers/crates/compilers/src/flatten.rs
```diff
@@ -222,12 +231,19 @@ impl Flattener {
...
+            has_complete_ast_build(&output.sources, &ordered_sources, build_id)
+        });
 
         // Convert all ASTs from artifacts to strongly typed ASTs
         let mut asts: Vec<(PathBuf, SourceUnit)> = Vec::new();
         for (path, ast) in output.sources.0.iter().filter_map(|(path, files)| {
-            if let Some(ast) = files.first().and_then(|source| source.source_file.ast.as_ref())
+            let source = target_build_id
+                .as_ref()
+                .and_then(|build_id| files.iter().find(|source| source.build_id == *build_id))
```
mablr: If the target build has an entry here with `ast: None`, it shadows a valid AST from another build; if it has no entry, `files.first()` can mix invocation-local AST IDs. Since pruning is merely skipped, missing ASTs can also leave imports in the flattened output. Could we select a build only when it has ASTs for every ordered source, then use that build exclusively?

stevencartavia: [f2ca2c2](https://github.com/foundry-rs/foundry-core/pull/125/commits/f2ca2c2e03ba8e07d31b7437f3d51771be12d9f6)

crates/compilers/crates/compilers/tests/project.rs
```diff
@@ -599,6 +599,188 @@ fn test_flatteners(project: &TempProject, target: &Path, additional_checks: fn(&
     additional_checks(&result);
 }
 
+// <https://github.com/foundry-rs/foundry/issues/7239>
+#[test]
+fn can_flatten_named_import_through_intermediary() {
```
mablr, line 604: **Optional:** Could we extend this coverage with a multi-hop renamed re-export (for example, `A -> {Renamed} from B -> {Original as Renamed} from C`)? That would exercise the recursive declaration-owner traversal plus aliasing at more than one intermediary. Non-blocking for this PR.

crates/compilers/crates/compilers/src/flatten.rs
```diff
@@ -881,3 +1012,61 @@ pub fn combine_version_pragmas(pragmas: &[impl AsRef<str>]) -> Option<String> {
...
+                }),
+            },
+            version: Version::new(0, 8, 30),
+            build_id: build_id.to_string(),
+            profile: "default".to_string(),
+        }
+    }
+
+    #[test]
+    fn complete_ast_build_ignores_incomplete_target_build() {
```
mablr, line 1047: **Optional:** This directly covers build selection well. A follow-up integration test through `Flattener::new`, where the first target `VersionedSourceFile` has no AST but a later coherent build does, would also guard the wiring between compiler output and this helper. Non-blocking.

crates/compilers/crates/compilers/src/flatten.rs
```diff
@@ -222,12 +222,23 @@ impl Flattener {
             sources
         };
 
-        let sources = Source::read_all(&ordered_sources)?;
+        let mut sources = Source::read_all(&ordered_sources)?;
+
+        // AST IDs are only unique within a single compiler invocation. Select a target build only
```
mablr, line 227: **FYI / follow-up:** `Flattener::new` currently requests AST output by updating only `project.settings` above, while `Project::update_output_selection` updates every settings profile. If flattening can select a non-default profile, it may be worth adding coverage and switching to the project-level helper separately. This appears pre-existing, not introduced by this PR.

mablr (approved):
LGTM

I left some optional follow-up ideas

mattsse (approved):
LGTM

mattsse (approved):

crates/compilers/crates/compilers/tests/project.rs
stevencartavia: https://github.com/foundry-rs/foundry-core/pull/126

crates/compilers/crates/compilers/src/flatten.rs
stevencartavia: https://github.com/foundry-rs/foundry-core/pull/127

crates/compilers/crates/compilers/src/flatten.rs
stevencartavia: deferring this one

PR #128, opened 2026-07-24
feat(output-selection): add `OutputSelection::empty()` and `insert` builder method

RyanLaidlaw:
## Motivation

`OutputSelection` currently only offers constructors that start from a broad default (`default_output_selection`, `complete_output_selection`, `common_output_selection`, `ast_output_selection`). There is no way to start from a truly empty selection  and build it incrementally. This would be useful when you want fine-grained control over what is requested, and this PR provides a simple builder pattern API to do just that.

`common_output_selection` is close but always uses `"*"` as the file key, so it can't express per-file selections.

## Changes

- Add `OutputSelection::empty()`, returning `Self(BTreeMap::new())`.
- Add `OutputSelection::insert(file, contract, outputs)`, a  builder method that inserts or extends the requested outputs for a given `(file, contract)` pair. Repeated calls for the same key append to the existing outputs rather than replacing them (not deduped).
- Doc comments with runnable examples for both.
- Unit tests covering:
  - building a selection across multiple files/contracts and asserting the
    exact serialized JSON,
  - merging outputs when `insert` is called twice for the same
    `(file, contract)` key.

Claude was consulted for feature validity and code correctness verification.

### DIFF
--- a/crates/compilers/crates/artifacts/solc/src/output_selection.rs
+++ b/crates/compilers/crates/artifacts/solc/src/output_selection.rs
@@ -70,6 +70,35 @@ impl OutputSelection {
         .into()
     }
 
+    /// Returns a completely empty output selection: `{}`.
+    /// 
+    /// Used for fine-grained control over which outputs are requested.
+    /// Combine with [`Self::insert`] to build the selection of outputs incrementally.
+    /// 
+    /// # Example
+    /// ```no_run
+    /// use foundry_compilers_artifacts_solc::output_selection::OutputSelection;
+    /// 
+    /// // Request only `abi` and `storageLayout`, for every contract in Counter.sol.
+    /// let selection = OutputSelection::empty()
+    ///     .insert("Counter.sol", "*", ["abi".to_string(), "storageLayout".to_string()]);
+    /// ```
+    pub fn empty() -> Self {
+        Self(BTreeMap::new())
+    }
+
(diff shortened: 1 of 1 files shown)


crates/compilers/crates/artifacts/solc/src/output_selection.rs
```diff
@@ -70,6 +70,35 @@ impl OutputSelection {
...
+    /// let selection = OutputSelection::empty()
+    ///     .insert("Counter.sol", "*", ["abi".to_string(), "storageLayout".to_string()]);
+    /// ```
+    pub const fn empty() -> Self {
+        Self(BTreeMap::new())
+    }
+
+    /// Inserts the set of requested outputs for a given `file` and `contract`
+    /// selector, returning `self` for chaining.
+    pub fn insert(mut self, file: impl Into<String>, contract: impl Into<String>, outputs: impl IntoIterator<Item = String>) -> Self {
```
mablr: Consuming method named `insert` seems weird, especially since it appends and preserves duplicates. Builder pattern may be more appropriate here?

RyanLaidlaw: I agree with the name. I updated it from `insert` to `with_output`. I'm ok with adding a new type `OutputSelectionBuilder(OutputSelection)` with the `with_output` method and a `build` method, but just wanted to check before.

mattsse (approved):
LGTM

PR #129, opened 2026-07-26
refactor(wallets): use tempo-alloy Accounts wallet

gakonst:
## Summary

Replace the parallel Tempo access-key config and signer tuple with tempo-alloy `TempoAccountsWallet`. Wallet resolution now returns one value that owns the root account, access-key signer, chain binding, and optional authorization, backed by the canonical Tempo Accounts `store.json`.

This removes the Foundry-specific `keys.toml`, SQLite channel database, TOML persistence, and their dependencies while retaining the no-`tempo` API shape through `MaybeTempoWallet`.

Pinned to tempoxyz/tempo#6965 at `9081cc65940c419ab60ebd1fd8170483ee49cfe1`; normal wallet builds avoid tempo-alloy node-only Reth/Revm dependencies.


gakonst:
Merging as I wanna iterate fast on this

PR #137, opened 2026-07-30
fix(compilers): preserve source unit names

stevencartavia:
Preserves exact Solidity source-unit names in raw compiler output instead of collapsing path-equivalent names into a single `PathBuf` entry. When projection is lossy, build-info retains the complete source and contract records with matching alias content, preserving cross-source AST references and allowing OpenZeppelin upgrades validation to consume Foundry build-info correctly.

Addresses foundry-rs/foundry#9299.

### DIFF
--- a/crates/compilers/crates/compilers/src/compilers/mod.rs
+++ b/crates/compilers/crates/compilers/src/compilers/mod.rs
@@ -289,6 +289,9 @@ impl<E, C> CompilerOutput<E, C> {
     }
 
     pub fn merge(&mut self, other: Self) {
+        // A build-info payload describes one compiler invocation and cannot represent merged
+        // outputs. Clear it before changing the projected sources and contracts.
+        self.build_info = None;
         self.errors.extend(other.errors);
         self.contracts.extend(other.contracts);
         self.sources.extend(other.sources);
@@ -324,6 +327,36 @@ impl<E, C> Default for CompilerOutput<E, C> {
             sources: BTreeMap::new(),
             metadata: BTreeMap::new(),
             build_info: None,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{BuildInfoPayload, CompilerOutput};
+    use std::collections::BTreeMap;
+
+    #[test]
+    fn merging_outputs_clears_build_info_payload() {
+        let payload = || {
+            Some(Box::new(BuildInfoPayload {
+                input: serde_json::Value::Null,
+                output: serde_json::Value::Null,
+                source_id_to_path: BTreeMap::new(),
+            }))
+        };
+
+        for (left, right) in [(true, false), (false, true), (true, true)] {
+            let mut output = CompilerOutput::<(), ()> {
+                build_info: left.then(payload).flatten(),
+                ..Default::default()
+            };
+            let other =
+                CompilerOutput { build_info: right.then(payload).flatten(), ..Default::default() };
+
+            output.merge(other);
+
+            assert!(output.build_info.is_none());
         }
     }
 }
--- a/crates/compilers/crates/compilers/src/compilers/mod.rs
+++ b/crates/compilers/crates/compilers/src/compilers/mod.rs
@@ -253,6 +253,22 @@ pub struct CompilerOutput<E, C> {
     pub sources: BTreeMap<PathBuf, SourceFile>,
     #[serde(default, skip_serializing_if = "::std::collections::BTreeMap::is_empty")]
     pub metadata: BTreeMap<String, serde_json::Value>,
+    /// Immutable compiler-run provenance used when the generic path-based projection would
+    /// collapse distinct source unit names. This intentionally survives transformations of the
+    /// projected output so build-info remains lossless.
+    #[doc(hidden)]
+    #[serde(skip)]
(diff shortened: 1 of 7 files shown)


figtracer (dismissed):

crates/compilers/crates/compilers/src/compilers/solc/mod.rs
```diff
@@ -115,6 +107,95 @@ impl Compiler for SolcCompiler {
...
+        let mut matches = original_sources
+            .iter()
+            .filter(|(input_name, _)| Path::new(input_name) == output_name.as_path());
+        let Some((_, source)) = matches.next() else { continue };
+        if matches.next().is_some() {
+            return Err(SolcError::msg(format!(
+                "multiple compiler input sources match output source unit `{output_name}`"
+            )));
+        }
+        input_sources.insert(output_name.to_string(), source.clone());
```
mablr: **Blocker:** This assumes path-equivalent aliases have identical content. A preprocessor can mutate the JSON input while solc loads the missing alias from disk, leaving build-info source text inconsistent with the alias AST. Can we resolve/read the exact alias using solc’s base/include-path rules, and error if resolution is ambiguous?

crates/compilers/crates/compilers/src/compilers/mod.rs
```diff
@@ -253,6 +253,22 @@ pub struct CompilerOutput<E, C> {
     pub sources: BTreeMap<PathBuf, SourceFile>,
     #[serde(default, skip_serializing_if = "::std::collections::BTreeMap::is_empty")]
     pub metadata: BTreeMap<String, serde_json::Value>,
+    /// Immutable compiler-run provenance used when the generic path-based projection would
+    /// collapse distinct source unit names. This intentionally survives transformations of the
+    /// projected output so build-info remains lossless.
+    #[doc(hidden)]
+    #[serde(skip)]
+    pub build_info: Option<Box<BuildInfoPayload>>,
```
mablr, line 261: This payload can become stale after `CompilerOutput::merge`, which merges projected sources/contracts but leaves `self.build_info` unchanged. Could `merge` clear or explicitly reject lossless payloads?

crates/compilers/crates/compilers/src/compilers/solc/mod.rs
```diff
@@ -545,6 +633,83 @@ mod tests {
...
+            errors: Vec::new(),
+        };
+
+        let output = compiler_output(&input, output).unwrap();
+        assert_eq!(output.sources[Path::new("src/file.sol")].id, 2);
+        assert_eq!(output.sources[Path::new("generated.sol")].id, 3);
+
+        let build_info = RawBuildInfo::new(&input, &output, true).unwrap();
+        let input_sources = &build_info.build_info["input"]["sources"];
+        assert_eq!(input_sources["src//file.sol"], input_sources["src/file.sol"]);
```
mablr, line 709: Could this use a real solc regression where preprocessed input differs from the filesystem alias? This fixture hard-codes the byte-identical assumption and has no AST, so it would not catch inconsistent build-info offsets.

mablr (commented):
One blocker and two suggestions

crates/compilers/crates/compilers/src/compilers/solc/mod.rs
stevencartavia: [6ae266b](https://github.com/foundry-rs/foundry-core/pull/137/commits/6ae266b8d464b39b9062263db9e80f97ce2857f7)

crates/compilers/crates/compilers/src/compilers/mod.rs
stevencartavia: [6ae266b](https://github.com/foundry-rs/foundry-core/pull/137/commits/6ae266b8d464b39b9062263db9e80f97ce2857f7)

crates/compilers/crates/compilers/src/compilers/solc/mod.rs
stevencartavia: [6ae266b](https://github.com/foundry-rs/foundry-core/pull/137/commits/6ae266b8d464b39b9062263db9e80f97ce2857f7)

mablr (dismissed):

crates/compilers/crates/compilers/src/compilers/solc/mod.rs
```diff
@@ -545,6 +634,94 @@ mod tests {
         },
     };
 
+    use super::{Solc, compiler_output};
+
+    fn solc() -> Solc {
+        Solc::new_with_version("solc", Version::new(0, 8, 26))
```
figtracer: this makes the new unit tests fail whenever `solc` on `PATH` is not exactly 0.8.26 because `new_with_version` debug-asserts the detected version. i reproduced `35 != 26` with solc 0.8.35; can this use the detected version or a non-validating test instance?

stevencartavia: [4e113c5](https://github.com/foundry-rs/foundry-core/pull/137/commits/4e113c5742ab377a288a097ba57528541746416c)

mablr (approved):

figtracer (approved):

PR #135, opened 2026-07-29
chore(deps): bump Tempo revision

decofe:
Updates the Tempo dependency to the unified Accounts and T10 revision from tempoxyz/tempo#6999.

Prompted by: @grandizzy

PR #138, opened 2026-07-30
fix(wallets): restore Tempo wallet API

mablr:
#121 replaced the public `MaybeTempoWallet` alias with a restricted wrapper. Although the return type name remained unchanged, this broke downstream consumers that rely on the complete `TempoAccountsWallet` API, including Foundry’s Cast and Forge integrations.

This restores `MaybeTempoWallet` as an alias for `TempoAccountsWallet`, preserves the existing chain ID configuration, and removes the dependencies used only by the partial wrapper. The wallet resolution test now explicitly checks the concrete return type to catch another incompatible substitution.

The account-scoping goal is valid, but foundry-wallets cannot implement it completely because `TempoAccountsWallet` owns account selection, access-key resolution, request preparation, raw signing, and authorization state. A follow-up in Tempo may need to add first-class account and chain scoping to that owning abstraction. Foundry-core can then consume the scoped behavior without replacing the concrete wallet API expected by downstream users.

The Touch ID implementation and its security hardening are unchanged.

### DIFF
--- a/crates/wallets/src/opts.rs
+++ b/crates/wallets/src/opts.rs
@@ -1,191 +1,15 @@
 #[cfg(feature = "tempo")]
 use crate::TempoAccountsWallet;
 use crate::{signer::WalletSigner, utils, wallet_raw::RawWalletOpts};
-#[cfg(feature = "tempo")]
-use alloy_consensus::Transaction;
-#[cfg(feature = "tempo")]
-use alloy_network::{NetworkWallet, TransactionBuilder};
 use alloy_primitives::Address;
-#[cfg(feature = "tempo")]
-use alloy_provider::{Provider, SendableTx, fillers::FillerControlFlow, fillers::TxFiller};
-#[cfg(feature = "tempo")]
-use alloy_transport::TransportResult;
 use clap::Parser;
 use eyre::Result;
 use serde::Serialize;
+
+/// When the `tempo` feature is enabled this is [`TempoAccountsWallet`];
+/// otherwise it is `()` so the API shape stays the same.
 #[cfg(feature = "tempo")]
-use tempo_alloy::{
-    TempoNetwork,
-    accounts::TempoAccessKey,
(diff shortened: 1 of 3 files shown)


figtracer (dismissed):

crates/wallets/src/opts.rs
```diff
@@ -354,7 +178,10 @@ impl WalletOpts {
             })?;
             let signer = utils::create_local_signer(access_key)?;
             let wallet = TempoAccountsWallet::from_secp256k1(root_account, signer, None);
-            let wallet = MaybeTempoWallet::new(wallet, root_account, chain_id);
+            let wallet = match chain_id {
+                Some(chain_id) => wallet.with_chain_id(chain_id),
```
stevencartavia, line 182: `maybe_signer_for_chain(4217)` no longer enforces that chain on the returned Tempo wallet. With `tempo_access_key` and `tempo_root_account` configured, I reproduced this through both `sign_request` and `sign_transaction_from`: the base rejects an explicit chain ID of `9999`, while this PR returns signed AA envelopes retaining `9999`. `TempoAccountsWallet::with_chain_id` defaults a missing request chain ID but does not reject explicit mismatches. Could we restore the mismatch validation, or update the “pinned to the selected chain” contract if default-only behavior is intended?

grandizzy:
will use this one for tempo bump too instead duping the effort

grandizzy:
Added the Tempo c72b8e98efeef1246a9b5f47f1f7b948259768af revision in 50bce74 so this PR carries both the dependency bump and its required wallet API compatibility fix. This supersedes #135.

crates/wallets/src/opts.rs
mablr: This is mentioned in pr description.

grandizzy (approved):

mattsse (approved):

PR #140, opened 2026-08-02
feat(wallets): expose Touch ID policy

dadadave80:
This change centralizes version-aware sidecar parsing and exposes public policy metadata so password rotation preserves `CurrentBiometry` for Touch ID-protected wallets. It addresses the [policy-preservation review on foundry-rs/foundry#15995](https://github.com/foundry-rs/foundry/pull/15995#discussion_r3696506069).

This PR was written primarily with Codex, including code generation, tests, and review assistance.


crates/wallets/src/touch_id/mod.rs
```diff
@@ -245,6 +245,11 @@ pub fn is_enrolled(keystore: &Path) -> bool {
     sidecar_path(keystore).exists()
 }
 
+/// Returns the access-control policy stored in the keystore's Touch ID sidecar.
+pub fn policy(keystore: &Path) -> Result<Policy, TouchIdError> {
+    Ok(read_sidecar(keystore)?.policy)
+}
```
mablr: This can be inlined

mablr (commented):
Makes sense

crates/wallets/src/touch_id/mod.rs
dadadave80: Added `#[inline]` in 366aa50.

mablr: I meant we can simply remove this function wrapper, not a big deal.

mablr (dismissed):

dadadave80:
Hi @stevencartavia could you please take a look at [foundry-core#140](https://github.com/foundry-rs/foundry-core/pull/140) when you have a moment? It exposes the persisted Touch ID policy so `wallet change-password` can preserve `CurrentBiometry` instead of silently resetting it to `UserPresence`. Does this address the concern you raised in your earlier review on [foundry#15995](https://github.com/foundry-rs/foundry/pull/15995#discussion_r3696506069)? Thanks

crates/wallets/src/touch_id/mod.rs
```diff
@@ -306,13 +306,7 @@ fn write_sidecar(keystore: &Path, sidecar: &Sidecar) -> std::io::Result<()> {
...
 }
 
-/// Unwraps the keystore password from the sidecar, triggering the enclave's
-/// access-control prompt (Touch ID under the default policy).
-///
-/// Blocks the calling thread until the prompt resolves; the authentication UI
-/// runs out of process, so calling from any CLI thread is fine. The returned
-/// password and the intermediate buffer are zeroed on drop.
-pub fn unwrap_password(keystore: &Path) -> Result<Zeroizing<String>, TouchIdError> {
+fn read_sidecar(keystore: &Path) -> Result<Sidecar, TouchIdError> {
```
stevencartavia, line 314: The current net diff no longer exposes a typed, version-aware accessor for the persisted policy: `policy()` was removed in `816a861`, while `read_sidecar()` and `Sidecar` remain private. Even after foundry#15995 updates its foundry-core revision, it would have to duplicate sidecar parsing to preserve the existing policy. Was dropping the public policy-reading API intentional? Restoring `policy()` would let foundry#15995 preserve the policy when it re-enrolls.

dadadave80: Thanks @stevencartavia, removing policy() was not intentional relative to this PR’s purpose leaving only a private read_sidecar() refactor and gives foundry#15995 no typed, version-aware way to preserve the enrollment policy.
My proposed correction is to restore policy(), keep Sidecar and read_sidecar() private, and update foundry#15995 to read the policy before rewriting the keystore and pass it back to enroll().
Hi @mablr, could you clarify whether you intended policy access to be removed entirely, or whether you would prefer a different public API such as a higher level re-enrollment operation that preserves the existing policy internally? I want to make sure the final API satisfies the downstream use case without exposing or duplicating the sidecar format. Thank you

mablr: Sorry, I thought it was a useless wrapper like agents often do.

Please restore `policy()` so Foundry can preserve the existing policy without parsing the sidecar itself.

dadadave80: Thanks for clarifying. I’ve restored `policy()` and its tests in 3e4c8f9738db5fb707476287b3a199653486acfe so Foundry can preserve the existing enrollment policy without depending on the private sidecar format.

mablr (approved):

stevencartavia (approved):

PR #145, opened 2026-08-03
fix(wallets): restore browser message signing

mablr:
Restore explicit browser-wallet signing methods for personal messages and dynamic EIP-712 typed data. Personal-message support was removed in https://github.com/foundry-rs/foundry/pull/13613 + https://github.com/foundry-rs/foundry/pull/13657.

Both methods use the existing browser signing request flow, parse wallet responses into Alloy signatures, and propagate wallet rejection, timeout, and malformed-signature errors.

This intentionally does not restore `Signer`, `SignerSync`, or `TxSigner` implementations for `BrowserSigner`. Browser transactions remain atomic sign-and-send operations, while downstream consumers can explicitly request personal-message or EIP-712 signing.

### DIFF
--- a/crates/wallets/src/wallet_browser/mod.rs
+++ b/crates/wallets/src/wallet_browser/mod.rs
@@ -14,8 +14,9 @@ mod types;
 mod tests {
     use std::time::Duration;
 
+    use alloy_dyn_abi::TypedData;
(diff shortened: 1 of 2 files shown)


figtracer (approved):

grandizzy (approved):
lgtm

PR #148, opened 2026-08-04
feat(compilers): preserve remapping ownership

stevencartavia:
Remapping autodetection currently collapses candidates globally by alias, discarding nested versions before callers can contextualize them.

This adds `RemappingDiscovery` and `Remapping::find_many_with_context`, retaining the preferred candidate for each lexical dependency owner and alias. It uses the existing traversal and priority rules, including scoped npm and symlink handling, while preserving the output and behavior of `Remapping::find_many`.

This allows consumers to contextualize duplicate transitive dependencies without rescanning dependency trees or reimplementing remapping discovery.

Part of OSS-583

### DIFF
--- a/crates/compilers/crates/artifacts/solc/src/remappings/find.rs
+++ b/crates/compilers/crates/artifacts/solc/src/remappings/find.rs
@@ -5,6 +5,7 @@ use std::{
     collections::{BTreeMap, HashSet, btree_map::Entry},
     fs::FileType,
     path::{Path, PathBuf},
+    sync::Mutex,
 };
 
 const DAPPTOOLS_CONTRACTS_DIR: &str = "src";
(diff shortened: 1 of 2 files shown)


crates/compilers/crates/artifacts/solc/src/remappings/find.rs
```diff
@@ -57,29 +57,17 @@ impl Remapping {
...
-                }
-            }
-        }
+        Self::find_many_with_context(dir).global
+    }
 
+    /// Attempts to autodetect global remappings and their nested dependency ownership.
+    ///
+    /// This performs the same traversal and global selection as [`Self::find_many`], while also
+    /// retaining the preferred nested candidate for each lexical owner and alias, using the same
```
figtracer: the shared symlink-target set drops one lexical owner when `a/lib/shared` and `b/lib/shared` both point to the same `shared/src/Shared.sol`; a local regression test returns only one context, nondeterministically. could cycle detection be scoped so both owner mappings survive?

figtracer (changes requested):
the ownership result is incomplete for shared symlink targets.

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
mablr: +1

stevencartavia: [c3b81ee](https://github.com/foundry-rs/foundry-core/pull/148/commits/c3b81eebd4d8268d328ce12567ff343f4cfcd11c)

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
```diff
@@ -57,45 +56,26 @@ impl Remapping {
...
-                    {
-                        e.insert(path);
-                    }
-                }
-                Entry::Vacant(e) => {
-                    e.insert(path);
-                }
-            }
-        }
+        Self::find_many_with_context(dir).global
```
mablr: `find_many` now performs ownership-aware traversal—including rescanning a shared symlink target per lexical owner—then discards the contextual output. Can we retain a global-only path (or cache/rebase candidates) so existing callers do not regress from roughly O(S + K) toward O(S × K)?

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
```diff
@@ -104,21 +84,77 @@ impl Remapping {
...
         }
 
-        all_remappings
+        let global = all_remappings
             .into_iter()
             .map(|(name, path)| Self { context: None, name, path: format!("{}/", path.display()) })
-            .collect()
+            .collect();
+        let contextual = contextual_remappings
+            .into_iter()
```
mablr, line 103: Could we account for nested contexts here? `BTreeMap` emits an ancestor before its descendant, while the project resolver uses the first match; this can select a parent dependency instead of solc’s longest-context winner.

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
```diff
@@ -104,21 +84,77 @@ impl Remapping {
...
+            .flat_map(|(context, remappings)| {
+                remappings.into_iter().map(move |(name, path)| Self {
+                    context: Some(format!("{}/", context.display())),
```
mablr, line 106: How should callers safely install this context? It is absolute, but `ProjectPathsConfig` compares a relative importer path; `into_relative` also drops the trailing separator, making `lib/a` match `lib/ab`. Please preserve the separator and provide a root-relative normalization path.

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
stevencartavia: [a379326](https://github.com/foundry-rs/foundry-core/pull/148/commits/a37932623b34eaa4b9e819626b9a90d76bdde59c)

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
stevencartavia: [a379326](https://github.com/foundry-rs/foundry-core/pull/148/commits/a37932623b34eaa4b9e819626b9a90d76bdde59c)

crates/compilers/crates/artifacts/solc/src/remappings/find.rs
stevencartavia: [a379326](https://github.com/foundry-rs/foundry-core/pull/148/commits/a37932623b34eaa4b9e819626b9a90d76bdde59c)

mablr (approved):
sgtm

figtracer (approved):

PR #154, opened 2026-08-10
feat(compilers): approve resolved executable paths

figtracer:
Adds approval-aware Solc and Vyper constructors that:

- resolve bare executable names through `PATH`
- canonicalize the selected executable before approval
- use that exact path for version probes and later compilation spawns
- fail closed when resolution fails

Includes regressions for bare-name resolution and symlink retargeting.

Companion to foundry-rs/foundry#16113.

### DIFF
--- a/crates/compilers/crates/compilers/src/compilers/compiler_path.rs
+++ b/crates/compilers/crates/compilers/src/compilers/compiler_path.rs
@@ -6,9 +6,37 @@ use std::{
     env,
     ffi::OsStr,
     path::{Path, PathBuf},
+    sync::{Arc, RwLock},
 };
 
-pub(super) fn resolve_and_approve(
+type CompilerApprovalHandler = dyn Fn(&Path) -> Result<()> + Send + Sync;
+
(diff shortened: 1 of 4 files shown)


crates/compilers/crates/compilers/src/compilers/compiler_path.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    let search_path = env::var_os("PATH").ok_or_else(|| not_found(path))?;
+    resolve_in_path(path, &search_path).ok_or_else(|| not_found(path))
+}
+
+fn resolve_in_path(path: &Path, search_path: &OsStr) -> Option<PathBuf> {
+    env::split_paths(search_path)
+        .flat_map(|dir| executable_candidates(dir.join(path)))
+        .find_map(|candidate| candidate.is_file().then(|| canonicalize(candidate).ok()).flatten())
```
mablr: This can select a non-executable file before a valid executable later in PATH. Could we match platform executable lookup semantics and add a regression for that ordering?

mablr (approved):

stevencartavia (approved):

PR #156, opened 2026-08-10
fix(wallets): reject invalid BIP-32 indices

stevencartavia:
Reject mnemonic derivation indices that set the BIP-32 hardened bit. This prevents out-of-range indices and derivation-path components from aliasing other wallet keys across raw and multi-wallet flows.

### DIFF
--- a/crates/wallets/src/signer.rs
+++ b/crates/wallets/src/signer.rs
@@ -29,6 +29,23 @@ use alloy_signer_turnkey::TurnkeySigner;
 
 pub type Result<T> = std::result::Result<T, WalletSignerError>;
 
+const BIP32_HARDEN: u32 = 0x8000_0000;
+
+fn validate_bip32_path(path: &str) -> Result<()> {
+    for component in path.split('/') {
+        let index = component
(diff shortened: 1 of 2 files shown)


mablr (dismissed):

crates/wallets/src/signer.rs
```diff
@@ -29,6 +29,23 @@ use alloy_signer_turnkey::TurnkeySigner;
 
 pub type Result<T> = std::result::Result<T, WalletSignerError>;
 
+const BIP32_HARDEN: u32 = 0x8000_0000;
+
+fn validate_bip32_path(path: &str) -> Result<()> {
```
mattsse: can we add this next to the other private fns in this file

stevencartavia: [c371b13](https://github.com/foundry-rs/foundry-core/pull/156/commits/c371b131ea3fb60dab8df9903fcb46356d18273d)

mablr (approved):

mattsse (approved):

PR #157, opened 2026-08-11
feat(fork-db): preserve exact fork ancestry

mablr:
Foundry can mix an initially resolved fork block with state or ancestry from its canonical replacement after a reorg.

This change pins state reads using EIP-1898 block hashes, derives `BLOCKHASH` ancestry by walking parent hashes, and separates caches by exact block and RPC source identity. Exact backends must be replaced rather than re-pinned, preventing stale requests from crossing fork identities.

This is the `foundry-fork-db` half of foundry-rs/foundry#15897. A companion Foundry PR carries the resolved identity through its fork consumers.


figtracer (dismissed):

crates/fork-db/src/backend.rs
```diff
@@ -468,27 +474,96 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
...
-                            Ok(KECCAK_EMPTY)
-                        }
-                        Err(err) => {
-                            error!(target: "backendhandler", %err, ?number, "failed to get block");
-                            Err(err)
+                    let block_hashes = async {
+                        if let Some(anchor) = anchor
+                            && number < anchor.number
+                        {
+                            eyre::ensure!(
```
stevencartavia: The companion integration uses `ForkBlock::with_rpc_number` for Nitro/Arbitrum with `l1BlockNumber` as the EVM-visible number and the L2 header number as `rpc_number`. When those differ, this guard makes an uncached `blockhash(block.number - 1)` fail with a database error. Could remapped-number chains use a compatible ancestry resolver or avoid selecting this ancestry path?

crates/fork-db/src/backend.rs
```diff
@@ -468,27 +474,96 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
...
+                                        KECCAK_EMPTY,
+                                    )]))
+                                }
+                                Err(err) => {
+                                    error!(target: "backendhandler", %err, ?number, "failed to get block");
+                                    Err(err)
+                                }
+                            }
+                        } else {
+                            Ok(BlockHashData::from_iter([(U256::from(number), KECCAK_EMPTY)]))
```
stevencartavia: After `vm.roll` advances an exact fork, a `BLOCKHASH` request that is in range relative to the live block number but above the immutable anchor returns nonzero `KECCAK_EMPTY`. For example, after rolling from `N` to `N+2`, `blockhash(N+1)` returns `0xc5d246…a470`. For this absent post-anchor height, could this return `B256::ZERO` so the opcode yields zero instead of a synthetic nonzero hash?

crates/fork-db/src/backend.rs
```diff
@@ -468,27 +474,90 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
...
-                        }
-                        Err(err) => {
-                            error!(target: "backendhandler", %err, ?number, "failed to get block");
-                            Err(err)
+                    let block_hashes = async {
+                        if let Some(anchor) = anchor
+                            && number < anchor.number
+                            && anchor.number == anchor.rpc_number
+                        {
+                            if anchor.number - number > 256 {
```
stevencartavia: The 256-block check is relative to the immutable anchor, not the live EVM block number. After `vm.roll` from block 1000 to 700, `blockhash(699)` is within revm’s live 256-block window, but this branch measures 301 blocks from the anchor and returns `KECCAK_EMPTY`. Reproduced through EVM execution. Could this anchor-relative cutoff be removed so the database continues walking the exact ancestry?

mablr: d654e3e

stevencartavia (dismissed):

crates/fork-db/src/backend.rs
```diff
@@ -468,27 +474,84 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
...
+                                    descendant.number - 1,
+                                    rpc_number,
+                                    header.parent_hash(),
+                                );
+                                hashes.insert(U256::from(descendant.number), descendant.hash);
+                            }
+                            Ok(hashes)
+                        } else if anchor.is_none_or(|anchor| anchor.number != anchor.rpc_number) {
+                            let block = provider
+                                .get_block_by_number(number.into())
```
figtracer, line 646: this still returns the wrong hash on arbitrum: `number` is the evm-visible l1 height, while `eth_getBlockByNumber` indexes l2 blocks. i reproduced the same height returning `0x183b…6051` while remote `blockhash` returned `0xdbb0…2f44`. can this use the remote-evm resolver from #159?

mablr: yes current was done earlier, we need to coordinate, I guess this is not a regression from the current diff.

mablr: It should be addressed with branch update https://github.com/foundry-rs/foundry-core/pull/157/commits/dbc6c6caeb3020312fb8f9e22c6c711cbc8270e7, I added a test to cover exact intersection https://github.com/foundry-rs/foundry-core/pull/157/commits/f8d75988007e3a0ab180470fff48ae3848f565da.

crates/fork-db/src/backend.rs
```diff
@@ -932,17 +1024,38 @@ impl<N: Network, B: ForkBlockEnv> SharedBackend<N, B> {
...
+        provider: P,
+        db: BlockchainDb<B>,
+        anchor: ForkBlock,
+    ) -> (Self, BackendHandler<N, B>) {
+        let (backend, backend_rx) = unbounded();
+        let cache = Arc::new(FlushJsonBlockCacheDB(Arc::clone(db.cache())));
+        db.block_hashes().write().insert(U256::from(anchor.number), anchor.hash);
+        let block_id = BlockId::from((anchor.hash, Some(false)));
+        let handler =
+            BackendHandler::new(provider.erased(), db, backend_rx, Some(block_id), Some(anchor));
```
stevencartavia, line 1051: `new_with_anchor` accepts a DB with unbound metadata or a `fork_hash` different from `anchor.hash`, so cached account/storage values can be served without any hash-pinned state RPC. This was reproduced for both cases; after flushing, `meta.fork_hash` remained A while `block_hashes[10]` contained anchor B. Could this return an error unless `fork_hash == Some(anchor.hash)` and `source_id.is_some()`?

stevencartavia (approved):

figtracer (approved):
makes sense

0xsarawut-eth (approved):

PR #158, opened 2026-08-11
fix(compilers): support Solar version probing

DaniPopes:
Solar emits Solc-compatible version metadata when `SOLC_WRAPPER=1` is set. The compiler version probes now set that variable for executables whose filename contains `solar`, in both synchronous and asynchronous paths. The parser then handles Solar and Solc output through the shared `Version:` format without Solar-specific parsing rules.

This PR was developed with AI assistance; the implementation and checks were reviewed by the author.

### DIFF
--- a/crates/compilers/crates/compilers/src/compilers/solc/compiler.rs
+++ b/crates/compilers/crates/compilers/src/compilers/solc/compiler.rs
@@ -732,34 +732,6 @@ fn parse_version(stdout: &str) -> Result<Version> {
     Ok(Version::from_str(&version.replace(".g++", ".gcc"))?)
 }
 
-#[cfg(test)]
-mod version_tests {
-    use super::*;
-
-    #[test]
-    fn parses_solc_version_output() {
-        let version = parse_version(
-            "solc, the solidity compiler commandline interface\n\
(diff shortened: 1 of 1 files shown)


crates/compilers/crates/compilers/src/compilers/solc/compiler.rs
```diff
@@ -711,71 +717,37 @@ fn compile_output(output: Output) -> Result<Vec<u8>> {
...
 
+fn parse_version(stdout: &str) -> Result<Version> {
+    let version = stdout
+        .lines()
+        .rev()
+        .find_map(|line| {
+            let line = line.trim();
+            line.strip_prefix("Version: ")
+        })
+        .ok_or_else(|| SolcError::msg("Version not found in Solc output"))?;
```
stevencartavia: This narrows version parsing to `Version: ...` lines, so `Solc::new` now fails version detection for executables that emit a bare semver—such as `solcjs`—even though the previous parser accepted this output. Would it be worth falling back to parsing the final nonempty line when no `Version: ...` line is found and adding a small regression test?

stevencartavia (approved):

decofe (approved):

PR #159, opened 2026-08-12
fix(fork-db): resolve Arbitrum block hashes in the EVM

stevencartavia:
Arbitrum uses L1 block numbers for `BLOCKHASH`, while `eth_getBlockByNumber` uses L2 block numbers. Resolve block hashes through a pinned `eth_call` on Arbitrum so ArbOS applies the correct mapping.


crates/fork-db/src/backend.rs
```diff
@@ -468,26 +509,45 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
...
-
-                    let block_hash = match block {
-                        Ok(Some(block)) => Ok(block.header().hash()),
-                        Ok(None) => {
-                            warn!(target: "backendhandler", ?number, "block not found");
-                            // if no block was returned then the block does not exist, in which case
-                            // we return empty hash
-                            Ok(KECCAK_EMPTY)
+                    let block_hash: eyre::Result<B256> = async {
+                        let cached_chain = meta.read().chain;
```
mablr, line 562: Could we invalidate legacy Arbitrum block-hash entries? Existing persisted hashes are returned before this chain check, so caches created by the old `eth_getBlockByNumber` path remain incorrect after upgrading.

crates/fork-db/src/backend.rs
```diff
@@ -468,26 +509,45 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
                 self.db.cache().record_cache_miss();
                 entry.insert(vec![listener]);
                 let provider = self.provider.clone();
+                let block_id = self.block_id;
```
mablr, line 557: The result now depends on `block_id`, but cache and in-flight deduplication are keyed only by `number`. After `set_pinned_block`, a cached or pending result from the old pin can be reused. Could the pin/generation be included in that identity?

crates/fork-db/src/backend.rs
```diff
@@ -468,26 +509,45 @@ impl<N: Network, B: ForkBlockEnv> BackendHandler<N, B> {
...
+                        let chain = if let Some(chain) = cached_chain {
+                            chain
+                        } else {
+                            let chain = Chain::from(
+                                provider.get_chain_id().await.wrap_err("failed to get chain ID")?,
+                            );
+                            meta.write().chain = Some(chain);
+                            chain
+                        };
+                        if chain.is_arbitrum() {
```
mablr: Does this intentionally exclude custom Orbit/Nitro chains? `is_arbitrum()` recognizes named Arbitrum networks only, so custom chain IDs still take the incorrect L2-number lookup path.

stevencartavia: [70f4e19](https://github.com/foundry-rs/foundry-core/pull/159/commits/70f4e19f8ef11da7510d9e216fd6730ad14c6b2e)

crates/fork-db/src/backend.rs
stevencartavia: [70f4e19](https://github.com/foundry-rs/foundry-core/pull/159/commits/70f4e19f8ef11da7510d9e216fd6730ad14c6b2e)

crates/fork-db/src/backend.rs
stevencartavia: [70f4e19](https://github.com/foundry-rs/foundry-core/pull/159/commits/70f4e19f8ef11da7510d9e216fd6730ad14c6b2e)

crates/fork-db/src/backend.rs
```diff
@@ -94,6 +96,13 @@ const ACCOUNT_FETCH_SUPPORTS_ACC_INFO: u8 = 1;
 /// Use regular individual getCode, getNonce, getBalance calls
 const ACCOUNT_FETCH_SEPARATE_REQUESTS: u8 = 2;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum BlockHashStrategy {
+    Unknown,
+    Rpc,
+    Evm,
+}
```
mablr: This feels a bit weird...

stevencartavia: [cda7757](https://github.com/foundry-rs/foundry-core/pull/159/commits/cda775711e60d481bee13456d239d50e4e9b46bf)

mablr (approved):

figtracer (approved):

PR #161, opened 2026-08-13
fix(compilers): match Windows contexts

mablr:
Fixes contextual remapping resolution for external dependencies on Windows. Mixed path separators are normalized before context matching, while source-unit paths covered by contextual remappings are made project-relative to avoid ambiguity with Windows drive-letter colons in Solidity remapping syntax. Other source-unit paths remain unchanged to preserve existing cache and artifact identities.


figtracer (dismissed):

crates/compilers/crates/compilers/src/compile/project.rs
```diff
@@ -431,16 +431,33 @@ impl<L: Language, S: CompilerSettings> CompilerSources<'_, L, S> {
...
+                    crate::utils::normalize_solidity_import_path(_root, &context).ok()
+                })
+                .collect::<Vec<_>>();
             self.sources.values_mut().for_each(|versioned_sources| {
                 versioned_sources.iter_mut().for_each(|(_, sources, _)| {
                     *sources = std::mem::take(sources)
                         .into_iter()
                         .map(|(path, source)| {
+                            let path = if contexts.iter().any(|context| path.starts_with(context)) {
+                                relative_path(&path, _root).unwrap_or(path)
```
stevencartavia: On Windows, this re-keys sources beneath a remapping context before cache and sparse-output filtering, while the graph and loaded cache retain absolute paths. This can cause unchanged sources to recompile and absolute-path sparse filters to produce zero artifacts. Could internal keys remain absolute, with conversion only at the compiler I/O boundary?

stevencartavia (approved):

figtracer (approved):

PR #179, opened 2026-09-06
fix(compilers): retry busy solc spawns

decofe:
Launching solc can fail with `Text file busy` even after an installer closes its writable handle before publishing the binary: a concurrently forked child can retain a duplicate until it execs. This complements [alloy-rs/svm-rs#230](https://github.com/alloy-rs/svm-rs/pull/230) and addresses the execution-side failure seen in [this Foundry CI job](https://github.com/foundry-rs/foundry/actions/runs/34018298207/job/101489061398?pr=16689).

Retry only `ExecutableFileBusy` failures from process creation in synchronous and asynchronous compilation and version detection. Five exponential delays total 310 ms; successful processes, compiler failures, and other I/O errors are never retried. Async retries yield to the runtime, and exhausted retries preserve the original path and I/O error.

All four launch-path regressions failed with OS error 26 before the patch and pass afterward. Linux tests publish an executable while a duplicate writable descriptor remains open, cover large compiler input and single-threaded async scheduling, and verify that compiler failures execute only once. A real solc 0.8.18 binary also passed all four recovery paths and compiled a Solidity contract. Twenty-five stress repetitions completed 50,000 compiler calls during 25,000 replacements with no failures; an unguarded control spawn failed with `ETXTBSY` on every replacement.

Workspace-wide nightly Clippy and formatting checks pass, as does `cargo check` with only the `async` feature enabled. The final `cargo test --workspace --all-features --no-fail-fast -- --skip external` run passed. The unfiltered suite needs unavailable explorer API credentials/paid access. Earlier runs had intermittent project-cache and warning-filter assertions; the warning-filter failures also reproduced on unchanged base [fccc0fe](https://github.com/foundry-rs/foundry-core/commit/fccc0fe4a3e69cd9b27c53adeb1fe7122a9a8b80), and the final rerun passed without changes to those tests or their code.

Independent review found that using Tokio timers would panic on an I/O-only runtime. An intermediate blocking-pool delay could instead deadlock when the only blocking worker drove the compiler future. Async retries now use an optional `futures-timer` delay, independent of both facilities; this lazily starts one shared timer thread on the first busy retry. Public-API regressions cover version detection and compilation without a timer driver, plus a saturated blocking pool with a bounded timeout. These regressions failed against the respective earlier implementations and pass with the final fix.

The independent-timer revision passed 14 focused nextest tests, 75 repeated runtime-edge tests with retries disabled, real solc 0.8.18 integration, the async-only check, nightly Clippy and formatting. A separate reviewer independently confirmed the deadlock reproducer now completes and all three runtime-edge regressions pass.

An earlier CI run also exposed an overly strong expectation in the adversarial stress test: 1,000 consecutive busy publications can outlast all six spawn attempts. The test now accepts only typed `ExecutableFileBusy` during that contention phase, still checks every successful output exactly, rejects all other errors, and requires unconditional version/compile success after all workers finish. An additional public-API regression verifies persistent-busy errors retain their kind and compiler path. These changes were independently reviewed; the bounded production retry policy is unchanged. Nextest retries are disabled for both spawn regression modules so runner retries cannot hide failures.

Final local validation: 15 focused regressions passed, followed by 394/394 workspace tests with all features and runner retries disabled. The existing default filter excluded 41 external tests; four other tests remained ignored (45 skipped total). All 25 additional adversarial publication stress repetitions passed with every worker and child process pinned to one CPU and runner retries disabled. Final nightly Clippy, formatting, and diff checks passed.

This PR was written and tested with AI assistance (Codex).

Prompted by: @DaniPopes

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -4,3 +4,8 @@
 retries = { backoff = "exponential", count = 2, delay = "5s", jitter = true }
 slow-timeout = { period = "30s", terminate-after = 3 }
 default-filter = "not test(/external_/)"
+
+# Retry regressions must pass on their first attempt, not be hidden by test-runner retries.
+[[profile.default.overrides]]
+filter = "test(compilers::solc::compiler::spawn_tests::) | test(compilers::solc::command::tests::)"
+retries = 0
--- a/crates/compilers/crates/compilers/Cargo.toml
+++ b/crates/compilers/crates/compilers/Cargo.toml
@@ -72,6 +72,7 @@ async = [
     "tokio/fs",
     "tokio/process",
     "tokio/io-util",
+    "tokio/time",
     "foundry-compilers-artifacts/async",
 ]
 # Enables `svm` to auto-detect and manage `solc` builds.
--- a/crates/compilers/crates/compilers/Cargo.toml
+++ b/crates/compilers/crates/compilers/Cargo.toml
@@ -32,6 +32,7 @@ path-slash.workspace = true
 yansi.workspace = true
 solar.workspace = true
 futures-util = { workspace = true, optional = true }
+futures-timer = { version = "3.0.4", optional = true }
 tokio = { workspace = true, optional = true }
 
 auto_impl = "1.3.0"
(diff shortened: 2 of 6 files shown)


DaniPopes (approved):

figtracer (approved):

mattsse (approved):
lgtm

PR #180, opened 2026-09-06
perf(compilers): reuse cached ABI output

figtracer:
Read-only ABI requests repeatedly compile sources missing from the normal artifact cache. `compile_abi_cached` reuses validated normal artifacts first, then acquires ABI misses through the existing compiler validation pipeline. Publish immutable generations atomically under a store lock; unavailable storage or contention returns in-memory output. Prune retired generations, legacy payloads and invalid source/settings contexts while preserving distinct valid filter contexts until invalidation or cleanup. Normal artifacts, compiler paths and build provenance remain intact, and failed secondary cleanup no longer prevents primary cleanup. This supplies https://github.com/foundry-rs/foundry/pull/16695. Measurements use profiling builds against Foundry master at `34ba6dec`, the local `foundry-bench` runner, and ten samples per binary in counterbalanced batches. These are warm-cache results; initial compilation, refresh copying, and contention are excluded. The full-cache control was repeated with explicit test warmups after a noisy first batch; its wall time is neutral and user CPU is 1.6% higher. Code and PR text were generated with Codex, with iterative Omp/Astra xhigh review.

### Results

| Workload | Master wall | PR wall | Wall change | User CPU, master → PR | Result on both |
| --- | ---: | ---: | --- | ---: | --- |
| Aave v4 `1e8de863`, `EngineFlagsTest::test_toBool_zero_returnsFalse`, partial cache | 4.706 s | 0.396 s | 11.89× faster | 4.498 → 0.507 s | 1 pass; gas 8,493 |
| Solady `acd959aa`, `SafeCastLibTest::testSafeCastUint256ToInt256Bench`, full cache | 138.04 ms | 137.92 ms | Neutral (within noise) | 210.81 → 214.11 ms | 1 pass; gas 29,446 |


mattsse (dismissed):
lgtm

0xKarl98 (dismissed):

crates/compilers/crates/compilers/src/compile/project.rs
```diff
@@ -340,21 +463,51 @@ impl<'a, T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
...
+                        compiled_artifacts,
+                        persist: false,
+                    });
+                }
+                Err(err) => return Err(err),
+            };
 
             // emits all the build infos, if they exist
-            output.write_build_infos(project.build_info_path())?;
+            if let Err(err) = output.write_build_infos(&cache.storage_paths().build_infos) {
```
mablr, line 537: The artifact-before-manifest ordering exists in the primary cache, but failures there propagate. This new optional-storage path instead swallows build-info/manifest write errors after `on_output` has overwritten stable artifact paths. The old secondary manifest can therefore point at new artifact contents, and a later source restoration can return a false hit with the wrong ABI/provenance. Could secondary writes use generation-specific paths and publish the manifest only after the whole generation succeeds?

crates/compilers/crates/compilers/src/compile/project.rs
```diff
@@ -340,21 +463,51 @@ impl<'a, T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
...
                 &output.contracts,
                 &output.sources,
-                &project.paths,
+                cache.storage_paths(),
                 ctx,
                 &primary_profiles,
-            )?;
+            ) {
+                Ok(artifacts) => artifacts,
+                Err(err) if optional_storage && matches!(err, SolcError::Io(_)) => {
```
mablr, line 517: This fallback misses some filesystem failures from `ArtifactFile::write`: `create_parent_dir_all` currently converts them to `SolcError::Message`. For example, a regular file at `artifacts/Contract.sol` makes ABI acquisition fail instead of returning the in-memory output. Please preserve/catch the filesystem classification and add a blocked nested-artifact-path test.

crates/compilers/crates/compilers/src/compile/project.rs
```diff
@@ -247,6 +248,121 @@ impl<'a, T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
...
+                                })
+                                .collect::<Vec<_>>();
+                            (language.to_string(), version, profile, files)
+                        })
+                    })
+                    .collect::<Vec<_>>();
+                jobs.sort_unstable();
+                let mut mocks = normal_mocks.iter().collect::<Vec<_>>();
+                mocks.sort_unstable();
+                let identity = serde_json::to_vec(&(jobs, mocks))?;
```
mablr: This identity includes jobs and mocks, but not the preprocessor implementation/configuration; additionally, an all-normal-cache hit returns before this identity is checked. Since `Preprocessor` may mutate input based on the complete job or its own configuration, the same files/profile can reuse output produced under different semantics. Can reuse require an explicit preprocessing identity/context in both tiers, or be disabled when one is unavailable?

crates/compilers/crates/compilers/src/lib.rs
```diff
@@ -350,6 +350,15 @@ impl<T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler> Pro
...
     /// ```
     pub fn cleanup(&self) -> std::result::Result<(), SolcIoError> {
+        let abi_cache = self.abi_cache_path();
+        if abi_cache.exists() {
+            if abi_cache.is_dir() {
+                std::fs::remove_dir_all(&abi_cache)
+            } else {
+                std::fs::remove_file(&abi_cache)
+            }
+            .map_err(|err| SolcIoError::new(err, &abi_cache))?;
```
mablr: An error deleting the optional ABI cache now returns before the primary manifest/artifacts/build-info are cleaned. In Foundry, `--force` drops cleanup warnings, so an undeletable `.abi` subtree can leave the normal cache reusable and defeat forced recompilation. Please attempt primary cleanup even if secondary cleanup fails, then return the saved error.

crates/compilers/crates/compilers/src/compile/project.rs
```diff
@@ -247,6 +248,121 @@ impl<'a, T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
...
+                directory.push(foundry_compilers_core::utils::unique_hash(identity));
```
mablr: Each distinct preprocessed job/source set creates a persistent context, while only full project cleanup removes old ones. As users vary `--match-*` filters, contexts that normal compilation has made unnecessary remain on disk, including their artifacts and build contexts. Is that unbounded lifecycle intentional, or should obsolete contexts be pruned/capped?

crates/compilers/crates/compilers/tests/project.rs
```diff
@@ -141,6 +143,810 @@ fn can_compile_hardhat_sample() {
...
+        }
+    }
+    let mut project = TempProject::<MultiCompiler>::dapptools().unwrap();
+    project.set_solc("0.8.30");
+    project.project_mut().update_output_selection(|selection| {
+        *selection = OutputSelection::common_output_selection(["abi".to_string()]);
+    });
+    for name in ["First", "Second", "Third"] {
+        project
+            .add_source(name, &format!("pragma solidity ^0.8.0; contract {name} {{}} "))
```
mablr: This currently fails both Clippy jobs with `needless_borrows_for_generic_args`. Please pass `format!(...)` directly instead of `&format!(...)`.

crates/compilers/crates/compilers/src/cache/abi.rs
```diff
@@ -0,0 +1,202 @@
...
+        let _ = fs::remove_dir_all(directory.join("artifacts"));
+        let _ = fs::remove_dir_all(directory.join("build-info"));
+        // Only the lock holder can stage or publish; no reader can be loading retired files.
+        if let Ok(entries) = fs::read_dir(directory) {
+            for entry in entries.flatten() {
+                let path = entry.path();
+                let name = entry.file_name();
+                let name = name.to_string_lossy();
+                if path != published && name.starts_with("generation-") {
+                    let _ = fs::remove_dir_all(path);
```
mablr, line 154: `ProjectCompileOutput` can outlive this lock and exposes `ArtifactFile.file` paths into the generation. A later refresh deletes those paths here, so a caller retaining the first output can see `NotFound` after an unrelated second compile. If these paths are intentionally transient, could `compile_abi_cached` document that callers must consume the in-memory artifacts? The “no reader” comment currently sounds stronger than the public output lifetime.

DaniPopes (dismissed):

0xKarl98 (dismissed):

crates/compilers/crates/compilers/src/cache/abi.rs
figtracer: thats fine

mablr (approved):

mattsse (approved):
lgtm

PR #183, opened 2026-09-07
chore(ci): migrate to shared CI checks

decofe:
Migrate CI from `tempoxyz/ci` to `tempoxyz/gh-actions`.

- Use shared Rust lint in deny-only mode, with nightly, all features, `--locked`, and the StepSecurity secret mapping.
- Route all third-party actions through vendored or first-party `gh-actions` replacements across all five workflows; keep GitHub-owned actions directly SHA-pinned.
- Add workflow scanning and Pinact validation, with an exception for full-SHA `gh-actions` references, and require scanning in the CI success gate.
- Commit `Cargo.lock` with dependencies resolved for the seven-day cooldown; gate Cargo dependency-consuming CI and nightly jobs on the strict cooldown check and use `--locked`. Retain Foundry's exact `wnaf@0.14.1` exception.
- Address browser-wallet workflow lint findings: group output writes, document the guarded `workflow_run`, and explicitly retain checkout credentials for the existing bot-branch push.

The vendored taiki installer selects nextest `0.9.140` instead of `0.9.143`. Toolchains and standalone tools are not age-checked; existing test matrices are unchanged.

Workflow and lockfile changes were authored with Centaur AI assistance.

Prompted by: @grandizzy


.pinact.yaml
```diff
@@ -0,0 +1,10 @@
+version: 3
+
+rules:
+  # Tempo's shared workflows are SHA-pinned but do not publish releases that
+  # pinact can use for version-comment or minimum-age verification.
+  - ignore: true
+    conditions:
+      - expr: |
+          ActionRepoFullName == "tempoxyz/gh-actions"
+      - expr: ActionVersion matches "^[0-9a-f]{40}$"
```
mablr: Pinact ORs separate `conditions`, so this exempts every full-SHA action from minimum-age validation. Combine both predicates in one expression with `&&` so only SHA-pinned `tempoxyz/gh-actions` references are ignored.

.github/workflows/ci.yml
```diff
@@ -219,12 +230,21 @@ jobs:
...
+    uses: tempoxyz/gh-actions/.github/workflows/rust-lint.yml@9b0664503f535261c1d4af0aadf0098bb7cbf26f # main
     permissions:
       contents: read
+    secrets:
+      STEP_SECURITY_API_KEY: ${{ secrets.STEP_SECURITY_API_KEY }}
+    with:
+      run-clippy: false
+      run-fmt: false
+      run-typos: false
+      deny-rust-toolchain: nightly
```
mablr, line 246: The shared workflow installs nightly only on the host; cargo-deny still runs in its container’s default Rust 1.85 because `rust-version` is not forwarded. Please pass `rust-version: ${{ inputs.deny-rust-toolchain }}` in `tempoxyz/gh-actions`, then repin this workflow and verify the container reports nightly.

.pinact.yaml
mablr: https://github.com/foundry-rs/foundry-core/pull/183/commits/3378091155f5ef94d4b150c4395dda037c0d35c3

.github/workflows/cargo-cooldown.yml
mablr, line 1: Do we actually want cooldown here ?

Foundry has it, wouldn't this be a duplicate? Moreover, it adds Cargo.lock, which is weird for lib?

mablr (dismissed):

.github/workflows/ci.yml
mablr: https://github.com/foundry-rs/foundry-core/pull/183/commits/bcc24775656cf78e24c8d8ae18219387264a7460

figtracer (approved):

grandizzy (approved):

mablr (approved):

PR #190, opened 2026-09-09
fix(fork-db): omit requireCanonical from anchored block id

### LINKED ISSUE #16769: anvil: anchored fork sends requireCanonical:false, rejected by RSKj/Rootstock (-32603)
### Component

Anvil

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### What version of Foundry are you on?

anvil 1.8.1-stable

### What version of Foundryup are you on?

N/A

### What command(s) is the bug in?

 anvil --fork-url <rootstock rpc url> --fork-block-number <N>

### Operating System

Linux

### Describe the bug

Since v1.8.0, an anchored fork (`--fork-block-number`) pins every state read to the fork block hash using an EIP-1898 object with an explicit `"requireCanonical": false`:

```json
{"jsonrpc":"2.0","id":1,"method":"eth_getTransactionCount",
 "params":["0x8a25d8c9fa8c7a726137f2d618d85cbc2c083f78",{"blockHash":"0x7115…","requireCanonical":false}]}
```

RSKj (Rootstock, `RskJ/9.0.3`) deserializes the block object into a `Map<String, String>` and throws on the boolean, returning `-32603 Internal server error`. anvil then aborts the fork:

```
failed to get account for 0x8a25d8c9fa8c7a726137f2d618d85cbc2c083f78: server returned an error response: error code -32603: Internal server error
```

The same fork works on foundry ≤ 1.7.1, which read state by block number.

The pinning comes from `SharedBackend::new_with_anchor` in foundry-core (`crates/fork-db/src/backend.rs`):

```rust
let block_id = BlockId::from((anchor.hash, Some(false)));
```

`requireCanonical` defaults to `false` per EIP-1898, so sending it explicitly adds nothing. Passing `None` makes alloy serialize the bare hash string, which geth, reth and RSKj all accept and which keeps reads pinned to the exact hash.

Proposed fix: https://github.com/foundry-rs/foundry-core/pull/190

dkatzan:
## Motivation

Since the exact block anchoring introduced in #157 (shipped in foundry v1.8.0), every state read made by an anchored `SharedBackend` (`eth_getBalance`, `eth_getTransactionCount`, `eth_getCode`, `eth_getStorageAt`, `eth_call`) is pinned with `BlockId::from((anchor.hash, Some(false)))`, which alloy serializes as:

```json
{"blockHash": "0x…", "requireCanonical": false}
```

Some RPC implementations do not accept `requireCanonical` as a JSON boolean. RSKj (Rootstock) deserializes the EIP-1898 object into a `Map<String, String>` and fails with `-32603 Internal server error` when the value is a boolean. As a result `anvil --fork-url <rootstock> --fork-block-number N` aborts on the first account fetch:

```
failed to get account for 0x…: server returned an error response: error code -32603: Internal server error
```

The same fork works on foundry ≤ 1.7.x, which read state by block number.

## Solution

Pass `None` for `require_canonical` in `SharedBackend::new_with_anchor`. alloy then serializes the block id as a bare 32-byte hash string (`"0x…"`) instead of the EIP-1898 object. This is the form geth has always accepted as a block parameter ([rpc/types.go](https://github.com/ethereum/go-ethereum/blob/master/rpc/types.go#L196)), and alloy/reth parse it the same way, so state is still read strictly by the anchor hash and the reorg protection from #157 is unchanged.

`requireCanonical` defaults to `false` per EIP-1898, so nothing is lost by not sending it explicitly.

The two anchor tests that asserted the explicit `"requireCanonical": false` object now assert the bare anchor hash.

The unrelated `Some(false)` in `block_hash_via_evm` (chain heuristic for `BLOCKHASH` opcode handling) is intentionally left untouched.

Fixes [16769](https://github.com/foundry-rs/foundry/issues/16769)

## AI disclosure

The root-cause investigation was done with Claude Code; the patch is a one-line change plus two test assertion updates, reviewed and tested manually (`cargo test -p foundry-fork-db`).

### DIFF
--- a/crates/fork-db/src/backend.rs
+++ b/crates/fork-db/src/backend.rs
@@ -1046,7 +1046,7 @@ impl<N: Network, B: ForkBlockEnv> SharedBackend<N, B> {
         let (backend, backend_rx) = unbounded();
         let cache = Arc::new(FlushJsonBlockCacheDB(Arc::clone(db.cache())));
         db.block_hashes().write().insert(U256::from(anchor.number), anchor.hash);
-        let block_id = BlockId::from((anchor.hash, Some(false)));
+        let block_id = BlockId::from((anchor.hash, None));
         let handler =
             BackendHandler::new(provider.erased(), db, backend_rx, Some(block_id), Some(anchor));
         Ok((Self { backend, cache, blocking_mode: Default::default(), exact: true }, handler))
@@ -1460,8 +1460,7 @@ mod tests {
             let mut storage_request = server.recv().unwrap();
             let storage: Request = serde_json::from_reader(storage_request.as_reader()).unwrap();
             assert_eq!(storage.method, "eth_getStorageAt");
-            assert_eq!(storage.params[2]["blockHash"], anchor_hash.to_string());
-            assert_eq!(storage.params[2]["requireCanonical"], false);
+            assert_eq!(storage.params[2], anchor_hash.to_string());
             storage_request
                 .respond(Response::from_string(
                     serde_json::json!({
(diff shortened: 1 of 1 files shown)


mattsse (approved):
oh, I see, yeah this should not be set like this, ty!

mablr (approved):

PR #192, opened 2026-09-10
chore(ci): replace Cargo cooldown with Socket

decofe:
Apply the per-job Socket setup from [foundry#16765](https://github.com/foundry-rs/foundry/pull/16765) to Rust CI, nightly tests, and both browser-wallet install/build jobs. Use the same pinned `secure-runner`/`socket-firewall` revision, connection-timeout workaround, unknown-host handling, Cargo shim check, and explicit sccache startup. Remove Cargo cooldown and run the existing cargo-deny version/checks on the host.

Keep caches, frozen pnpm lockfiles, and the guarded bundle-commit job unchanged. Run Linux test jobs on pull requests and retain the full nine-leg Linux/macOS/Windows test matrix on pushes to main, with the existing Rust versions and exclusions. Restrict nightly failure issue creation to main so branch runs can validate safely. Preserve the working Linux compiler `wrap` rules and select host-only `bypass` rules for `binaries.soliditylang.org` and `raw.githubusercontent.com` on macOS/Windows, retaining `wrap` for explorer test APIs. The raw GitHub bypass covers the entire host. Disable Cargo certificate revocation checking for the Windows test leg while retaining certificate validation, and keep all five Socket/Cargo settings explicit in Clippy; no Foundry-specific fixtures or report uploads. Crate-age policy and exceptions now belong in Socket, including the previous revm/wnaf exceptions. As in Foundry, cached packages and direct binary downloads are not rechecked; fork PRs without OIDC use Socket Free without organization policies.

The full cross-platform CI run passed after rerunning one macOS job that hit a Socket connection timeout. The temporary PR test-matrix expansion has been reverted; CI for the restored PR matrix is pending.

Workflow changes were written with AI assistance.

Prompted by: @grandizzy


.github/workflows/browser-wallet.yml
```diff
@@ -25,35 +25,43 @@ jobs:
...
-          version: 11.9.0
-          run_install: false
       - name: Set up Node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version-file: crates/wallets/src/wallet_browser/app/src/.nvmrc
+      - name: Secure Linux runner
+        uses: tempoxyz/gh-actions/actions/secure-runner@5338a3746a2ac2ddd88cbede733c79f907aca3a0
+        with:
+          egress-policy: audit
```
sds: No need to specify this explicitly—[it is the default](https://github.com/tempoxyz/gh-actions/blob/5338a3746a2ac2ddd88cbede733c79f907aca3a0/actions/secure-runner/action.yml#L13).

Specifying it also prevents us from changing enforcement from audit to blocking in future via StepSecurity dashboard policy.

.github/workflows/browser-wallet.yml
```diff
@@ -25,35 +25,43 @@ jobs:
...
-        uses: tempoxyz/gh-actions/vendor/pnpm/action-setup@134b6c9105953bcf6d50ef0f3f70e31b7a3ec64f
-        with:
-          version: 11.9.0
-          run_install: false
       - name: Set up Node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version-file: crates/wallets/src/wallet_browser/app/src/.nvmrc
+      - name: Secure Linux runner
+        uses: tempoxyz/gh-actions/actions/secure-runner@5338a3746a2ac2ddd88cbede733c79f907aca3a0
```
sds, line 43: I think I understand why we need to put this after the setup-node action, but that seems like a serious bug in our `secure-runner` action. Will investigate.

grandizzy: @sds ptal this combo, think I got this right 
https://github.com/tempoxyz/gh-actions/pull/156
used in
https://github.com/foundry-rs/foundry-core/pull/195

PR #194, opened 2026-09-11
fix(fork-db): support required account-info reads

mablr:
Tempo's `eth_getBalance` returns a placeholder native balance, while `eth_getAccountInfo` returns the real balance. Racing these methods can cache incorrect execution state: Relay's multicaller then sends an extra native refund, adding 13,836 gas and failing the canary introduced in https://github.com/foundry-rs/foundry/pull/16793.

Add `AccountFetchPolicy::RequireAccountInfo` so callers can require authoritative account data without fallback. The policy is part of cache identity, preventing reuse of incompatible caches even in offline-start mode. The default retains automatic account loading; the Foundry integration selects the new policy for Tempo forks.

AI assistance was used for implementation and validation.

### DIFF
--- a/crates/fork-db/README.md
+++ b/crates/fork-db/README.md
@@ -9,3 +9,10 @@ Provides a shared, caching database layer backed by a remote RPC provider, allow
 ## Features
 
 - `zstd`: Enables [zstd](https://github.com/gysber/zstd-rs) compression support.
+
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

figtracer (approved):

PR #201, opened 2026-09-18
fix(compilers): invalidate changed imports

figtracer:
Changing a remapping can alter a source's resolved dependency without changing any Solidity file content. If both the old and new target already have cached artifacts, the compiler currently reports an unchanged build and reuses bytecode linked to the old target.

This compares each cache entry's persisted resolved imports with the freshly resolved graph and invalidates the source when they differ. The regression test keeps both targets warm, retargets one remapping, and verifies that the dependent source recompiles against the new target.

This discrepancy was found by the dynamic test-linking differential cache campaign in foundry-rs/foundry#16920, but it also reproduces with preprocessing disabled.

AI assistance was used to investigate the cache path and implement this change.


grandizzy (approved):
lgtm

mablr (approved):
👍

PR #202, opened 2026-09-18
fix(compilers): track native preprocessing dependencies

mablr:
Dynamic test linking removes contract creation bytecode from tests and scripts, allowing dependency implementation changes to reuse their cached artifacts. Some deployments cannot be rewritten, however, so their creation bytecode remains embedded and must still invalidate consumers.

This adds a cache contract for preprocessors to report exact or conservative native bytecode dependencies. These classifications are persisted and merged across sparse compiler jobs, profiles, and compiler versions, with conservative state taking precedence. They are intentionally retained while artifacts for the source may survive; this favors safe over-invalidation over stale bytecode reuse.

The cache also records the preprocessor policy version and source-unit context so incompatible preprocessing decisions cannot reuse prior state. This provides the foundry-core side of the fixes for [foundry#16682](https://github.com/foundry-rs/foundry/issues/16682) and [foundry#16901](https://github.com/foundry-rs/foundry/issues/16901). The corresponding Foundry integration reports the dependencies that preprocessing leaves native.

AI-assisted.


crates/compilers/crates/compilers/src/cache.rs
```diff
@@ -1275,6 +1471,66 @@ impl<'a, T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
...
+    }
+
+    /// Applies classifications from compiled jobs, replacing only their matching contexts.
+    pub fn apply_native_dependency_updates(&mut self, updates: NativeDependencyUpdates) {
+        let ArtifactsCache::Cached(cache) = self else { return };
+        for ((version, profile), (processed, dependencies)) in updates.contexts {
+            for file in processed {
+                let versions = cache.cache.native_dependencies.entry(file.clone()).or_default();
+                let profiles = versions.entry(version.clone()).or_default();
+                profiles.remove(&profile);
```
grandizzy: **[P1] Preserve native dependencies when the corresponding artifact survives**

`processed` includes sources passed to preprocessing as `SourceCompilationKind::Optimized`, whose artifacts are not regenerated. Removing their previous classification here can discard dependencies still required by the cached bytecode.

A failing sequence:
1. A full build caches `Consumer` with embedded `Dep` creation bytecode and records `Consumer → Dep`.
2. An importing test changes. `Consumer` is processed as an optimized import in a narrower compiler job, where preprocessing rewrites its deployment and reports no native dependencies.
3. This clears `Consumer → Dep`, but the rewritten artifact is not emitted—the original native bytecode remains cached.
4. An implementation-only change to `Dep` no longer invalidates `Consumer`, leaving stale embedded bytecode.

I reproduced this locally with a job-sensitive preprocessor: incremental bytecode differs from a clean build. Restoring only the discarded classification makes the comparison pass.

Could we replace classifications only for sources whose artifacts are actually regenerated, using the per-job `actually_dirty` set, and preserve or conservatively merge classifications for surviving artifacts?

*AI-assisted review; reproduction executed locally.*

grandizzy (changes requested):
Found a reproducible stale-bytecode case when dependency classifications are replaced for optimized imports whose artifacts remain cached. Details and suggested fix are in the inline comment.

AI-assisted review; reproduction executed locally.

crates/compilers/crates/compilers/src/cache.rs
mablr: https://github.com/foundry-rs/foundry-core/pull/202/commits/997ca0aee3517138a7bcaed3f2c0840d138e7ff9

crates/compilers/crates/compilers/src/compile/project.rs
```diff
@@ -128,18 +129,285 @@ use std::{
...
+    ///
+    /// Repeated updates within one compiler job are merged, with
+    /// [`NativeDependencyState::Conservative`] dominating. The cache applies the resulting
+    /// classification as a replacement for this compiler version and profile only.
+    /// Returns whether this was the source's first update in the request.
+    pub fn update(&mut self, file: PathBuf, state: Option<NativeDependencyState>) -> bool {
+        let first_update = self.processed_sources.insert(file.clone());
+        if !self.active_replacements.contains(&file) {
+            return first_update;
+        }
```
grandizzy, line 269: **[P1] Merge new dependency reports for optimized imports instead of discarding them**

This fixes the previous classification-clearing issue, but ignoring all reports for optimized imports introduces the reverse case. Their standalone artifacts survive, yet their freshly preprocessed code can still be embedded in a regenerated importer.

A failing sequence:
1. Initially, `Consumer` has no native dependency on `Dep`.
2. Changing `Caller` recompiles it with `Consumer` as an optimized import. In this narrower compiler job, preprocessing leaves `new Dep()` native in `Consumer`, and `Caller` embeds `Consumer` through `new Consumer()`.
3. The preprocessor reports `Caller → Consumer` and `Consumer → Dep`, but this guard discards the latter because `Consumer` is not in `actually_dirty`.
4. An implementation-only change to `Dep` no longer invalidates `Caller`, leaving its embedded bytecode stale.

I reproduced this with a synthetic job-sensitive preprocessor and real Solc compilation. The incremental-versus-clean bytecode comparison fails with the guard and passes when only this guard is disabled. Conversely, the original dependency-clearing reproduction passes with the guard and fails without it.

Could we replace classifications for regenerated artifacts, but preserve existing classifications **and conservatively merge new observations** for optimized imports? That retains protection for surviving bytecode without losing dependencies embedded in freshly generated consumers.

*AI-assisted review; reproduction executed locally.*

grandizzy (changes requested):
The original dependency-clearing issue is fixed, but the new guard introduces the reverse stale-bytecode case: newly discovered native dependencies of optimized imports are discarded even when their preprocessed code is embedded in a regenerated importer. Details and suggested correction are in the inline comment.

Existing suites passed serially: 94 unit tests, 108 project integration tests, and 6 mocked tests; one test ignored. The additional regression reproduction fails on this head.

AI-assisted review; reproduction executed locally with a synthetic job-sensitive preprocessor and real Solc compilation.

crates/compilers/crates/compilers/src/cache.rs
```diff
@@ -1275,6 +1496,88 @@ impl<'a, T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
...
+    ///
+    /// Recompiled artifacts replace their matching contexts. Observations from optimized inputs
+    /// are merged because newly compiled consumers can embed their transitive native dependencies.
+    pub fn apply_native_dependency_updates(&mut self, updates: NativeDependencyUpdates) {
+        let ArtifactsCache::Cached(cache) = self else { return };
+        for ((version, profile), updates) in updates.contexts {
+            for file in updates.processed {
+                let versions = cache.cache.native_dependencies.entry(file.clone()).or_default();
+                let profiles = versions.entry(version.clone()).or_default();
+                profiles.remove(&profile);
```
grandizzy: **[P2] Preserve merged observations while importing artifacts still survive**

Both earlier cases are fixed, but observations from optimized inputs are persisted in the same classification that this replacement removes. Regenerating the source's standalone artifact does not necessarily regenerate every cached importer that embeds its previously observed native dependencies.

A failing sequence:
1. Compiling `Caller` records `Caller → Consumer → Dep`, with `Consumer → Dep` coming from an optimized-import observation. `Caller`'s artifact embeds the native deployment.
2. Delete `Consumer`'s artifact and partially rebuild it in a compiler-job context where preprocessing removes its native deployment, keeping `Caller` out of scope.
3. This replacement clears `Consumer → Dep`, while `Caller`'s artifact and cache entry survive unchanged.
4. An implementation-only edit to `Dep` now incorrectly reuses `Caller`'s stale embedded bytecode.

I reproduced this with a synthetic job-sensitive preprocessor and real Solc compilation. The incremental-versus-clean bytecode comparison fails; restoring only the discarded observation makes the control pass. This requires a missing artifact, a partial rebuild, and job-dependent preprocessing. I have not demonstrated the exact sequence with Foundry's production preprocessor or established when this extended failure sequence first became possible.

Could we retain merged observations beyond standalone artifact replacement, or invalidate dependent cached artifacts before clearing them? Their lifetime needs to cover the importing artifacts that relied on them, not only the source artifact being regenerated.

*AI-assisted review; reproduction executed locally.*

grandizzy (changes requested):
Both previously reported reproductions now pass. One narrower correctness gap remains: rebuilding an imported artifact can clear observations still needed by a surviving cached importer. Details are in the inline comment.

Existing suites passed serially: 94 unit tests, 111 project integration tests, and 6 mocked tests; one test ignored. The additional partial-rebuild reproduction fails, while restoring only the discarded observation makes its control pass.

This is a remaining gap verified on this head, not a confirmed newly introduced regression. The reproduction uses a synthetic job-sensitive preprocessor and real Solc compilation; this exact sequence has not been demonstrated with Foundry's production preprocessor.

AI-assisted review; reproduction executed locally.

grandizzy (dismissed):
lgtm

figtracer (approved):

0xKarl98 (approved):

